PR #78, opened 2011-11-01
EOFError handling fix

danp:
We've been seeing things that talk to AWS sometimes get stuck and consume 100% CPU.

Tracked it down to how `EOFError` is handled. In this particular case there was no way for `Socket#read` to return `nil` after filling the buffer as much as possible and then getting `EOFError`. So [this loop](https://github.com/geemus/excon/blob/f245f613a3bc284dae346d0a8d567c7e523c6a1e/lib/excon/response.rb#L56) never stopped looping.

Not immediately sure how else something like this could be tested without resorting to the "bad" test server I added; let me know if there's a better way. But that server could be expanded to cover other things that are hard to test with the rackups, such as read timeouts.

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -11,6 +11,7 @@ module Excon
     def initialize(params = {}, proxy = {})
       @params, @proxy = params, proxy
       @read_buffer, @write_buffer = '', ''
+      @eof = false
 
       @sockaddr = if @proxy
         ::Socket.sockaddr_in(@proxy[:port].to_i, @proxy[:host])
@@ -39,6 +40,8 @@ module Excon
     end
 
     def read(max_length=nil)
+      return nil if @eof
+
       begin
         if max_length
           until @read_buffer.length >= max_length
@@ -64,6 +67,7 @@ module Excon
           raise(Excon::Errors::Timeout.new("read timeout reached"))
         end
       rescue EOFError
+        @eof = true
       end
       if max_length
         @read_buffer.slice!(0, max_length)


Paxa:
Having same issue sometimes randomly. I'm sending request to other rails application running Thin

geemus:
@Paxa do you have any backtraces or other info you could share?

Paxa:
```
Excon::Errors::SocketError: end of file reached (EOFError)
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/socket.rb:45:in `read_nonblock'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/socket.rb:45:in `readline'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/response.rb:53:in `parse'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/middlewares/response_parser.rb:6:in `response_call'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/connection.rb:372:in `response'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/connection.rb:236:in `request'
/Users/pavel/.rvm/gems/ruby-2.1.5/gems/excon-0.45.3/lib/excon/connection.rb:297:in `get'
```

I saw same errors many times in production too.

I added `p buffer` inside `def readline`
```ruby
    def readline
      return legacy_readline if RUBY_VERSION.to_f <= 1.8_7
      buffer = ''
      begin
        buffer << @socket.read_nonblock(1) while buffer[-1] != "\n"
        p buffer
```

Usually I receive something like this
```
"HTTP/1.1 200 OK\r\n"
"Content-Type: application/json\r\n"
"Content-Length: 132\r\n"
"ETag: \"a2b1dd21e457293703e9b430fa73be11\"\r\n"
"Cache-Control: max-age=0, private, must-revalidate\r\n"
"X-Request-Id: 656c124e-0760-4689-903d-426defccda8f\r\n"
"X-Runtime: 0.019141\r\n"
"Connection: keep-alive\r\n"
"Server: thin\r\n"
"\r\n"
```

But when error happen - no print for response, means it happen when it start reading first line.

I'm using persistent connection with https://github.com/mperham/connection_pool, usually it happen when I send request (create connection and keep in pool), then wait few minutes and try to use that connection again.

In that case this error happen. Request actually not sent (server not receive it), after error, if I try to send request again it works fine (connection recovered)

geemus:
Yeah, that matches with behavior I've seen. Basically the remote server responds with `keep-alive`, but has a timeout set such that inactive connections will still be closed after some timeous (30 seconds seems pretty common). When that happens, the remote end of the socket has been closed, but the local side doesn't really know that. So it naively tries to write onto the socket and things blow up. Are you able to do this with `idempotent` turned on to retry? Not sure if that will retry in this case or not (if it doesn't perhaps we should consider modifying it so that it does).

PR #184, opened 2013-01-24
Use ruby's built-in resolver to get the host address

rubiojr:
Use the built-in Resolv class to get the host address before
passing it to Socket.getaddrinfo.

One of the things you can do with this is to replace the nameserver
the ruby process is using, and set different one, without having
to tweak your system's resolver settings (i.e. /etc/resolv.conf)
or manually resolving every host before passing it to an Excon request

Imaging we want to use a specific nameserver to resolv non public
domain names, available in our company's private nameserver:

```
require 'resolv'
require 'excon'

external_dns = '192.168.1.1'
resolver = Resolv::DNS.new(:nameserver => [external_dns], :ndots => 1)
Resolv::DefaultResolver.instance_variable_set :@resolvers,
                                              [resolver]
Excon.get 'https://myserver.mycompany.lan'
```

Where 'myserver.mycompany.lan' is known only to the DNS server at
192.168.1.1. We could also use resolv-replace.rb from the stdlib
and transparently replace the resolver globally to our process.

We also use this kind of stuff to run tests using different nameservers
transparently, without having to update system's configuration. i.e.:

```
external_dns = ENV['NAMESERVER']
resolver = Resolv::DNS.new(:nameserver => [external_dns], :ndots => 1)
Resolv::DefaultResolver.instance_variable_set :@resolvers,
                                              [resolver]

Tests.run # test code here
```

And then using the env variable before running the tests:

```
$ NAMESERVER=192.168.1.1 rake test
```

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -37,9 +37,11 @@ module Excon
       exception = nil
 
       addrinfo = if @data[:proxy]
-        ::Socket.getaddrinfo(@data[:proxy][:host], @data[:proxy][:port], @data[:proxy][:family], ::Socket::Constants::SOCK_STREAM)
+        host = Resolv.getaddress @data[:proxy][:host]
+        ::Socket.getaddrinfo(host, @data[:proxy][:port], @data[:proxy][:family], ::Socket::Constants::SOCK_STREAM)
       else
-        ::Socket.getaddrinfo(@data[:host], @data[:port], @data[:family], ::Socket::Constants::SOCK_STREAM)
+        host = Resolv.getaddress @data[:host]
+        ::Socket.getaddrinfo(host, @data[:port], @data[:family], ::Socket::Constants::SOCK_STREAM)
       end
 
       addrinfo.each do |_, port, _, ip, a_family, s_type|


oleander:
Any news on this one? I'm currently using Sidekiq with this library to fetch data from an external source and I'm experiencing freezes.

geemus:
No news that I'm aware of. I'm also not certain this would impact the particular issue you are seeing. Could you go into a bit more detail on what you are seeing and what led you to this issue in particular? Thanks!

bjeanes:
@geemus Well there's this as a decent reason to consider using the built in Resolv: http://blog.gregburek.com/2015/02/22/dns-eglibc-and-resolv-replace-on-heroku/

geemus:
@bjeanes nice, thanks. I just was totally missing some context I think.

@gregburek - any advice on approaching this? I guess we could probably just steal some of the bits from resolv-replace and add them in here. I just figured since you had dived deeper there you may insights/advice. Thanks!

gregburek:
@geemus resolve-replace is dangerous to use in a lib because it monkey patches the IPSocket layer for all requests that could use DNS. There are definitely gems that seem to rely on broken behavior (https://github.com/mperham/sidekiq/issues/1258#issuecomment-27389456) so this is not a good blanket solution. resolv-replace also seems to not do ipv6 or respect any host level dns tricks, such as /etc/hosts or system caches. The breadth of places where excon is run would suggest that custom DNS behavior may be a surprise as a default, but a legit work around for people fighting other problems.

geemus:
@gregburek thanks for the additional info and context. Definitely agree that it isn't something to move hastily upon, but wanted to better understand options. Thanks!

PR #449, opened 2014-12-05
Unwrap ipv6 addresses for use with sockets

jrafanie:
Literal IPv6 addresses, such as ::1, have a few rules in regards to their use:
* They must be wrapped in [], such as [::1] when used in URL/URIs and in the Host Request header field.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=45891
TL;DR: 
RFC 2617 says Host = host : port
RFC 2732 says, host must be wrapped in [], so, we need **Host = [::1]:9292**

* They must be unwrapped, "::1", when passed to sockets.
See: https://github.com/ruby/ruby/blob/e70210cad637c5be31709f65c0eb9603f8ec18a1/lib/uri/generic.rb#L234

#### To illustrate the problem that this pull request fixes:
```ruby
irb(main):001:0> require 'excon'
=> true
irb(main):002:0> Excon.new("http://[::1]:9292").request
Excon::Errors::SocketError: getaddrinfo: nodename nor servname provided, or not known (SocketError)
...
```

I get a getaddrinfo error **because Excon is sending [::1]** directly to the socket layer.

#### Using this PR's fixed excon:

```ruby
$ bundle console
Resolving dependencies...
irb(main):001:0> require 'excon'
=> false

irb(main):002:0> Excon.new("http://[::1]:9292").request
Excon::Errors::SocketError: Connection refused - connect(2) (Errno::ECONNREFUSED)
...
```
**Note 1**: Instead of getaddrinfo error, I get connection refused because I have nothing listening.

**Note 2**: I don't like borrowing from ruby's URI class, I'd rather let URI handle this...

**Note 3**: It's unfortunate I have to unwrap each time we pass to the socket but I couldn't find an easy way to handle the request host header separately from the socket host value.  The same data/datum is used for both.  In other libraries, you can keep things as URI objects and call hostname for unwrapped (`::1`) and host for wrapped (`[::1]`).  As mentioned above in note 2, this is desirable but unfortunately, excon would have to change in many places to keep the host/port/scheme/path, etc. in URI format and extract out the parts you need when needed.  If we did do this though, we could then call URI#hostname where I'm currently unwrapping the [] for use in the socket code.

See also:
https://github.com/nahi/httpclient/pull/176
https://github.com/rest-client/rest-client/pull/332
https://github.com/rest-client/rest-client/pull/333

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -149,16 +149,27 @@ module Excon
 
     private
 
+    # IPv6 addresses need to be unwrapped for socket connections:
+    #   host: "[::1]" => "::1"
+    # Borrowed from ruby URI's:
+    #   https://github.com/ruby/ruby/blob/deba55eb1a950b72788aa4cab10ccc032c1d37a7/lib/uri/generic.rb#L650
+    def unwrapped_host_for_socket(v)
+      /\A\[(.*)\]\z/ =~ v ? $1 : v
+    end
+
     def connect
       @socket = nil
       exception = nil
 
       if @data[:proxy]
         family = @data[:proxy][:family] || ::Socket::Constants::AF_UNSPEC
-        args = [@data[:proxy][:host], @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
+
+        host = unwrapped_host_for_socket(@data[:proxy][:host])
+        args = [host, @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
       else
         family = @data[:family] || ::Socket::Constants::AF_UNSPEC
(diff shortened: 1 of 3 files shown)


Fryguy:
@jrafanie Would it make sense to change to the code to allow either a String *or* a URI, and warn a deprecation if the String comes in?  That way, you can let the tests/testing tell you where any additional fixes are needed, and if anything is missed it at least won't break.

geemus:
@Fryguy that definitely sounds like it could be a helpful intermediary step.

jrafanie:
Yeah, @Fryguy @geemus, sounds reasonable.  Maybe I'll do that first and fix any deprecations that get spit out in the test suite and then grep for any remaining instances after.

jrafanie:
Closing in favor of #459

PR #455, opened 2015-01-16
Re-implement timeout monitoring with IO.select

starbelly:
- decompose write into write_nonblock and write_block
 - decompose read into read_nonblock and read_block
 - added timeout_reached() and raise_timeout_error() to make monitoring DRY

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -10,6 +10,7 @@ module Excon
       Excon.display_warning('Excon::Socket#params is deprecated use Excon::Socket#data instead.')
       @data
     end
+
     def params=(new_params)
       Excon.display_warning('Excon::Socket#params= is deprecated use Excon::Socket#data= instead.')
       @data = new_params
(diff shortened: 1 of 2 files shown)


starbelly:
Right, so I know you said you had problems with IO.select in the past, so you may not want this code. None the less, I figured I'd initiate this convo with a PR, and at at the very leas I'm quite curious to see the CI results.

starbelly:
So no good for 1.8.7 and no good for ree. Will see what I can do. Do you have a target EOL date on support for 1.8.7?

geemus:
I haven't really thought particularly long nor hard on 1.8.7 drop. Probably would coincide with 1.0. Which also doesn't have a specific date per se. I feel like it should be fairly safe these days to drop it.

geemus:
Travelling this week for work, so a bit swamped, will try to come back and review this when I have a bit more bandwidth.

lib/excon/ssl_socket.rb
```diff
@@ -1,9 +1,8 @@
 module Excon
   class SSLSocket < Socket
-
-    HAVE_NONBLOCK = [:connect_nonblock, :read_nonblock, :write_nonblock].all? {|m|
+    HAVE_NONBLOCK = [:connect_nonblock, :read_nonblock, :write_nonblock].all? do|m|
```
geemus: Could you add a whitespace between do and | here?

lib/excon/ssl_socket.rb
```diff
@@ -85,12 +84,12 @@ def initialize(data = {})
       end
 
       if @data[:proxy]
-        request = 'CONNECT ' << @data[:host] << port_string(@data.merge(:omit_default_port => false)) << Excon::HTTP_1_1
+        request = 'CONNECT ' << @data[:host] << port_string(@data.merge('omit_default_port' => false)) << Excon::HTTP_1_1
```
geemus: Why not a symbol here?

lib/excon/ssl_socket.rb
```diff
@@ -101,7 +100,7 @@ def initialize(data = {})
         @socket.write(request)
 
         # eat the proxy's connection response
-        Excon::Response.parse(self, { :expects => 200, :method => "CONNECT" })
+        Excon::Response.parse(self,  'expects' => 200, 'method' => 'CONNECT')
```
geemus: Same here, symbols seem appropriate since they will appear frequently so we can not have to reallocate.

lib/excon/socket.rb
```diff
@@ -24,119 +25,44 @@ def initialize(data = {})
...
-            else
-              raise(Excon::Errors::Timeout.new("read timeout reached"))
-            end
-          end
-        rescue EOFError
-          @eof = true
-        end
+    def read(max_length = nil)
+      # if @eof
+      return max_length ? nil : '' if @eof
```
geemus: It is kind of a minor/preference thing, but could you decompose these into multi-line ifs instead of ternaries? Thanks!

geemus:
Nice! Overall looks really good and cleaner than my previous attempts. I have a few small comments above. Otherwise it looks like it passed on all the builds! Does this fix the timeout related issues you were seeing in-the-wild also? Thanks!

lib/excon/socket.rb
starbelly: Totally. I get OCD about condense conditionals.

lib/excon/ssl_socket.rb
starbelly: Will do

lib/excon/ssl_socket.rb
starbelly: I would love to have a symbol instead, but the travis CI builds were not passing with symbols in place here.

lib/excon/ssl_socket.rb
starbelly: See previous comment, it's a backwards compat issue. I prefer symbols.

starbelly:
Yeah, it definitely does the trick. We also get a slight performance increase using IO.select. Now, we need to talk about the caveat with using IO.select. That is the 1024 file descriptor limit that comes with using select(2) . Now that number isn't fixed ( i.e., depends on OS and can be circumvented via ulimit and such ). The best solution is not possible, since Ruby doesn't have a way to interface with poll(2) right now. The only other option is to write a native extension for handling sockets, which I'm not opposed to taking on that task, but is it necessary? I suppose the real question at hand what are/were the limitations of using Timeout.timeout ( aside from issues with green threads )?

geemus:
Timeout.timeout also just doesn't seem to work in many cases, for whatever reason. The symbols part is puzzling, I think they were symbols previously, so I'm not sure what would have changed to make it not work... I guess I'd prefer to try and dig deeper a bit there, but not the end of the world.

For clarity, how would the 1024 file descriptor issues manifest?

starbelly:
On the symbols, it might not have been the symbol, just lack of a rocket op between key and value. I'll have to try that again. 

 Anyway, on the FD limit issue : https://stackoverflow.com/questions/726352/poll-in-ruby 
Differing opinions there, and if you google this subject you'll find even more differing opinions. Basically, if someone uses excon and has over 1024 connections open, then tries to open one more funny things might start to happen, since select it limited to 1024 FD. I tried this on my macbook and got to 2000, after that I get thread errors, but that could be unrelated to select(). The way I'm trying to stress test this may not be suitable : 

```
require 'fog'
threads = 2100.times.map do
  Thread.new do
    puts "New fog!"
  connection = Fog::Storage.new(options)
  end
end
threads.each(&:join)

```
 Long story short, I'm not sure if this something we have to worry about or not. It seems to me the implementation of Timeout.timeout imposed more of limitation than select() will.

geemus:
Ah ha. Yeah, rockets are needed in the older rubies. If you could give that another shot I would appreciate it (hopefully it will be pretty quick/easy anyway).

Yeah, thanks for the details on limits. I guessed it was something like that, but figured having a more solid understanding would help (in case related issues end up coming up later).

Agreed that this seems reasonable, ie very few people are likely to approach this limit whereas (at least anecdotally) a number of people had issues with the timeouts not actually timing out as expected for various reasons.

So, overall looking good. If you can give the symbols one more try that would be great and then we can see about getting this in. Thanks again for all your hard work and patience as we discuss and work through this.

starbelly:
Tidied up last night ( symbols, whitespace, etc. ).

lib/excon/socket.rb
```diff
@@ -229,6 +149,112 @@ def connect
...
+                @read_buffer << @socket.read_nonblock(max_length - @read_buffer.length)
+              end
+            else
+              loop do
+                @read_buffer << @socket.read_nonblock(@data[:chunk_size])
+              end
+            end
+          rescue OpenSSL::SSL::SSLError => error
+            if error.message == 'read would block'
+              timeout_reached('read') ? raise_timeout_error('read') : retry
```
geemus: One, hopefully final thing. I think I'd prefer for these ternaries to be expanded also. Otherwise it's looking spot on. Thanks!

starbelly:
Forgot to comment that I tidied up some more ( i.e., decompose ternaries ).

geemus:
Ah, yes. Thanks for the ping (GH doesn't automagically ping on commits in PRs, though I kind of wish it did).

geemus:
Thanks!

PR #457, opened 2015-01-20
Tests and small simplify Excon::Utils

rusikf:
1 ) Add tests for methods in ```Excon::Utils```
2 ) Simplify it, bang methods is not needed, ```dup``` also not needed

### DIFF
--- a/lib/excon/utils.rb
+++ b/lib/excon/utils.rb
@@ -64,27 +64,21 @@ module Excon
 
     # Escapes HTTP reserved and unwise characters in +str+
     def escape_uri(str)
-      str = str.dup
       str.force_encoding('BINARY') if FORCE_ENC
-      str.gsub!(UNESCAPED) { "%%%02X" % $1[0].ord }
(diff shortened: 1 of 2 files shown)


lib/excon/utils.rb
```diff
@@ -64,27 +64,21 @@ def split_header_value(str)
 
     # Escapes HTTP reserved and unwise characters in +str+
     def escape_uri(str)
-      str = str.dup
       str.force_encoding('BINARY') if FORCE_ENC
```
dpiddy: Without `dup`ing won't this modify `str`?

rusikf: It looks, that ```force_encoding``` [doesn't change internal byte representation of the string](http://stackoverflow.com/questions/21614228/what-is-the-difference-betwee-encode-and-force-encoding-in-ruby)
```gsub``` method doen't modify str, only returned modified string

geemus:
Thanks!

PR #459, opened 2015-01-20
Pass ipv6 safe address to socket layer

jrafanie:
This pull request replaces #449 

The summary is that we need incoming valid URIs as Strings with IPv6 literals to be properly parsed for use at different layers.

For example: "http://[::1]:9292"
We need "::1" for Socket connections
We need "[::1]" for use in URIs and in the Request header's Host field

* In pull request #449, the code in Excon passes down "[::1]" for use everywhere, and the changes in the PR unwrap the address to "::1" for use with sockets.  The problem is that if we have need for "::1" elsewhere such as other socket connections, we need to repeat the process.  It's also duplicating URI#hostname's code.

* In this pull request, we instead pass along both URI#host ("[::1]") and URI#hostname ("::1") to Excon::Connection.new and ensure it's available for use by the Socket layer.  In this way, the Socket code can just use `@data[:hostname]` when it needs an unwrapped host value.  Any future socket code can also make use of this IPv6 safe hostname.

* Note, I researched changing Fog and the externalized fog plugins to pass URIs all over the place but it's a much harder problem and introduces much more risk into Fog and Excon.  This Pull request is much smaller and makes the valid assertion that if you pass a URI as a String, it should be a valid URI.

From the prior Pull Request #449:

Literal IPv6 addresses, such as ::1, have a few rules in regards to their use:
* They must be wrapped in [], such as [::1] when used in URL/URIs and in the Host Request header field.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=45891
TL;DR: 
RFC 2617 says Host = host : port
RFC 2732 says, host must be wrapped in [], so, we need **Host = [::1]:9292**

* They must be unwrapped, "::1", when passed to sockets.
See: https://github.com/ruby/ruby/blob/e70210cad637c5be31709f65c0eb9603f8ec18a1/lib/uri/generic.rb#L234

#### To illustrate the problem that this pull request fixes:
```ruby

irb(main):002:0> Excon.new("http://[::1]:9292").request
Excon::Errors::SocketError: getaddrinfo: nodename nor servname provided, or not known (SocketError)
...
```

#### Same problem in Fog itself:
```ruby
irb(main):001:0> Fog::Core::Connection.new("http://[::1]:9292").request({})
Excon::Errors::SocketError: getaddrinfo: nodename nor servname provided, or not known (SocketError)
```

I get a getaddrinfo error **because Excon is sending [::1]** directly to the socket layer.

```ruby
irb(main):006:0> TCPSocket.new("[::1]", 9292)
SocketError: getaddrinfo: nodename nor servname provided, or not known
```

#### Using this PR's fixed excon:

```ruby
irb(main):002:0> Excon.new("http://[::1]:9292").request
Excon::Errors::SocketError: Connection refused - connect(2) (Errno::ECONNREFUSED)
...
```
**Note 1**: Instead of getaddrinfo error, I get connection refused because I have nothing listening.

#### Finally, using this PR with Fog:
```ruby
irb(main):001:0> Fog::Core::Connection.new("http://[::1]:9292").request({})
Excon::Errors::SocketError: Connection refused - connect(2) for [::1]:9292 (Errno::ECONNREFUSED)
````

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -111,6 +111,7 @@ module Excon
       end
       params = {
         :host       => uri.host,
+        :hostname   => uri.hostname,
         :path       => uri.path,
         :port       => uri.port,
         :query      => uri.query,
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -11,6 +11,8 @@ require 'uri'
 require 'zlib'
 require 'stringio'
 
+require 'excon/extensions/uri'
+
 require 'excon/middlewares/base'
 require 'excon/middlewares/expects'
 require 'excon/middlewares/idempotent'
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -428,6 +428,7 @@ module Excon
           uri = @data[:proxy].is_a?(String) ? URI.parse(@data[:proxy]) : @data[:proxy]
           @data[:proxy] = {
             :host       => uri.host,
+            :hostname   => uri.hostname,
             # path is only sensible for a Unix socket proxy
             :path       => uri.scheme == UNIX ? uri.path : nil,
             :port       => uri.port,
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -35,7 +35,8 @@ module Excon
     #   @param [Hash<Symbol, >] params One or more optional params
     #     @option params [String] :body Default text to be sent over a socket. Only used if :body absent in Connection#request params
     #     @option params [Hash<Symbol, String>] :headers The default headers to supply in a request. Only used if params[:headers] is not supplied to Connection#request
-    #     @option params [String] :host The destination host's reachable DNS name or IP, in the form of a String
+    #     @option params [String] :host The destination host's reachable DNS name or IP, in the form of a String. IPv6 addresses must be wrapped (e.g. [::1]).  See URI#host.
+    #     @option params [String] :hostname Same as host, but usable for socket connections. IPv6 addresses must not be wrapped (e.g. ::1).  See URI#hostname.
     #     @option params [String] :path Default path; appears after 'scheme://host:port/'. Only used if params[:path] is not supplied to Connection#request
     #     @option params [Fixnum] :port The port on which to connect, to the destination host
     #     @option params [Hash]   :query Default query; appended to the 'scheme://host:port/path/' in the form of '?key=value'. Will only be used if params[:query] is not supplied to Connection#request
@@ -358,6 +359,12 @@ module Excon
         #params = params.dup
         #invalid_keys.each {|key| params.delete(key) }
       end
+
+      if validation == :connection && params.key?(:host) && !params.key?(:hostname)
+        Excon.display_warning('hostname is missing! For IPv6 support, provide both host and hostname: Excon::Connection#new(:host => uri.host, :hostname => uri.hostname, ...).')
+        params[:hostname] = params[:host]
+      end
+
       params
     end
 
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -155,10 +155,10 @@ module Excon
 
       if @data[:proxy]
         family = @data[:proxy][:family] || ::Socket::Constants::AF_UNSPEC
-        args = [@data[:proxy][:host], @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
+        args = [@data[:proxy][:hostname], @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
       else
         family = @data[:family] || ::Socket::Constants::AF_UNSPEC
-        args = [@data[:host], @data[:port], family, ::Socket::Constants::SOCK_STREAM]
+        args = [@data[:hostname], @data[:port], family, ::Socket::Constants::SOCK_STREAM]
       end
       if RUBY_VERSION >= '1.9.2' && defined?(RUBY_ENGINE) && RUBY_ENGINE == 'ruby'
         args << nil << nil << false # no reverse lookup
--- a/tests/servers/good.rb
+++ b/tests/servers/good.rb
@@ -333,6 +333,6 @@ end
 
 EM.run do
   EM.start_server("127.0.0.1", 9292, GoodServer)
-  EM.start_server("::1", 9293, GoodServer)
+  EM.start_server("::1", 9293, GoodServer) unless RUBY_PLATFORM == 'java'
   $stderr.puts "ready"
 end
--- a/tests/servers/good.rb
+++ b/tests/servers/good.rb
@@ -333,5 +333,6 @@ end
 
 EM.run do
   EM.start_server("127.0.0.1", 9292, GoodServer)
+  EM.start_server("::1", 9293, GoodServer)
   $stderr.puts "ready"
 end
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -73,6 +73,7 @@ module Excon
     :connect_timeout,
     :family,
     :host,
+    :hostname,
     :omit_default_port,
     :nonblock,
     :reuseaddr,
--- a/lib/excon/extensions/uri.rb
+++ b/lib/excon/extensions/uri.rb
@@ -1,0 +1,33 @@
+# TODO: Remove this monkey patch once ruby 1.9.3+ is the minimum supported version.
+#
+# This patch backports URI#hostname to ruby 1.9.2 and older.
+# URI#hostname is used for IPv6 support in Excon.
+#
+# URI#hostname was added in stdlib in v1_9_3_0 in this commit:
+#   https://github.com/ruby/ruby/commit/5fd45a4b79dd26f9e7b6dc41142912df911e4d7d
+#
+# Addressable::URI is also an URI parser accepted in some parts of Excon.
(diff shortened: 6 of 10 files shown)

### CODE REVIEW

**Comment by jrafanie** [comment]
@geemus Please review.   Thank you!

File: lib/excon/connection.rb [hunk 1]
```diff
@@ -358,6 +359,12 @@ def validate_params(validation, params)
         #params = params.dup
         #invalid_keys.each {|key| params.delete(key) }
       end
+
+      if validation == :connection && params.key?(:host) && !params.key?(:hostname)
+        Excon.display_warning('hostname is missing! For IPv6 support, provide both host and hostname: Excon::Connection#new(:host => uri.host, :hostname => uri.hostname, ...).')
```
jrafanie: @geemus will this warning cause issues by code outside of Excon?  Is Excon::Connection called directly or do libraries only call Excon.new?

lib/excon/socket.rb
```diff
@@ -155,10 +155,10 @@ def connect
 
       if @data[:proxy]
         family = @data[:proxy][:family] || ::Socket::Constants::AF_UNSPEC
-        args = [@data[:proxy][:host], @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
+        args = [@data[:proxy][:hostname], @data[:proxy][:port], family, ::Socket::Constants::SOCK_STREAM]
       else
         family = @data[:family] || ::Socket::Constants::AF_UNSPEC
-        args = [@data[:host], @data[:port], family, ::Socket::Constants::SOCK_STREAM]
+        args = [@data[:hostname], @data[:port], family, ::Socket::Constants::SOCK_STREAM]
```
jrafanie: These are the actual bug fixes.

jrafanie:
cc @fryguy @tenderlove @brandondunne @kbrock

lib/excon.rb
```diff
@@ -109,6 +109,7 @@ def new(url, params = {})
       end
       params = {
         :host       => uri.host,
+        :hostname   => uri.hostname,
```
jrafanie: All of the calls to URI#hostname will fail on ruby 1.9.2 and earlier because it wasn't added until 1.9.3.  I can add conditional logic around these calls if we want to support 1.9.2 and earlier.

jrafanie:
RE: The warning if you call Excon::Connection.new(:host => ...) but without a :hostname key...

The warning can be removed if need be, as I default hostname => host ( "[::1]") if it's missing but of course, ipv6 literals would fail with the reported `getaddrinfo: nodename nor servname provided, or not known (SocketError)`

Excon.new is not changed from the end user consumer perspective as it will now seamlessly deal with  wrapping [] or unwrapping the host value and passing the two values to Excon::Connection.new.

I looked through the readme and throughout the codebase and it doesn't look like Excon::Connection.new is advertised as being the public interface.  It looks like Excon.new is the recommended way.  Some tests explicitly call Excon::Connection.new and those have been updated in this PR.

I chose to change Excon#new since it already parsed the String URI to a URI object and handled merging in the Excon defaults.  I considered adding :hostname to the params hash passed to Excon::Connection.new to be of similar scope as merging defaults and validating the incoming URI String.

Based on any feedback in this area, I can also update the README to include a small paragraph to reflect the proper use of IPv6 literals with Excon.

Thanks!

lib/excon.rb
jrafanie: FYI, these calls are causing travis to fail on 1.8.7, 1.9.2 and ree.

jrafanie: I'm researching the jruby test failure.

tests/servers/good.rb
```diff
@@ -333,5 +333,6 @@ def split_header_value(str)
 
 EM.run do
   EM.start_server("127.0.0.1", 9292, GoodServer)
+  EM.start_server("::1", 9293, GoodServer)
```
jrafanie: So, on jruby, it's failing because I'm starting EventMachine and binding on  ::1:9293 and trying to connect to it in the test using http://[::1]:9293.  See the failure [here](https://travis-ci.org/excon/excon/jobs/47718169).  This works on MRI 1.9.3/2.0.0/2.1.5.

I couldn't find issues saying that you can't do this on jruby but I did find this:
http://stackoverflow.com/questions/15061805/connection-refused-thrift-client-on-jruby-1-7-3

Note:  I tried replacing ::1 and [::1] above to localhost and tests pass (but then ipv6 [] wrapping/unwrapping isn't being tested).  When requiring `tests/servers/good.rb` using localhost and `lsof` I see IPv6 file descriptors "listening" but again the purpose of the tests were to verify the correct usage of ::1 and [::1] for socket and non-socket usage.

I also tried "0:0:0:0:0:0:0:1" (the long form of ::1) and that also fails with `Connection refused - Connection refused (Errno::ECONNREFUSED)`

kbrock: I remember EventMachine being a little odd on jruby, but this may be fixed by now. And that error @jrafanie looks more like a potential issue in jruby. It is calling directly into java code.

Many places referencing jruby/jruby#775 when `::1` causes a jruby connection error.

jrafanie: Thanks @kbrock! 

I did notice that issue in my search yesterday.  I tried the latest eventmachine locally but still had the same problem.  I'll re-read the above issue and see if anything jumps out.  I was also going to try jruby 9000 to see if that also fails.

jrafanie:
@geemus I spent a while trying to isolate this but didn't make much progress.
Do you have any suggestions on dealing with rubies < 1.9.3? (ones no longer maintained)

For now, I disabled the IPv6 test on jruby.

Here is the "Connection Refused" errors when trying to connect to http://[::1]:9293 for the request_tests:
https://gist.github.com/jrafanie/b008dcd5b18207907559

Finally, here is the same tests run on jruby9000 pre1, it doesn't very far and fails early:
https://gist.github.com/jrafanie/dc107e9491f52206fb9d

Do you have any suggestions on this PR's direction?

lib/excon/connection.rb
geemus: In my experience other stuff usually uses `Excon.new` (I think). It is probably ok and I guess if it isn't we'll probably hear about it pretty quickly (and accidentally warning is not the end of the world).

lib/excon.rb
geemus: Yeah, if we could check ruby version and conditionally do it I think that would be preferred at least for the near future.

geemus:
Sorry for the delay, I was traveling for work last week and fell behind on OS stuff a bit. Overall this is looking good. I think if we can continue to support the older stuff and jruby for the time being that would be preferred though, so wrapping the bits in question with conditional logic would be great. Otherwise I think once travis is green we should probably be at least largely ready to go. Thanks!

jrafanie:
@geemus No worries on the delay... :bow:  Thanks for working on open source!  I've always feel horrible when maintainers can't seem to go on vacation or travel without checking in... 

@tenderlove offered the idea that it would probably be better to monkey patch URI#hostname into older rubies so it's really easy to remove later on when we no longer support 1.8.7 and 1.9.2 instead of having to add conditionals in 4 somewhat random places.

I'll try to get this green today, thanks again @geemus

geemus:
monkeypatching is kind of terrifying! But, it seems not too crazy in this context (as long as we make sure it only does it if it isn't already defined). Thanks!

jrafanie:
@geemus The PR has been updated to add commit c00f795 with the monkey patch.  I'm still waiting on travis to see if it passes.  I tested 1.9.2, 2.0.0 and 2.1.5 locally.  Hopefully, this monkey patch will be deleted sometime soon.

Thanks again!

jrafanie:
It's failing for something strange on ruby 2.2.
https://travis-ci.org/excon/excon/jobs/48515546

Seems like json 1.7.3 is failing to compile:
```
In file included from generator.c:1:0:
../fbuffer/fbuffer.h: In function ‘fbuffer_to_s’:
../fbuffer/fbuffer.h:158:47: error: macro "rb_str_new" requires 2 arguments, but only 1 given
../fbuffer/fbuffer.h:158:20: warning: initialization makes integer from pointer without a cast [enabled by default]
generator.c: In function ‘cState_aref’:
generator.c:632:5: warning: variable ‘state’ set but not used [-Wunused-but-set-variable]
generator.c: In function ‘isArrayOrObject’:
generator.c:867:5: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
make: *** [generator.o] Error 1
make failed, exit code 2
```

I amended my commit and force pushed so hopefully a rebuild on travis will be all green.

geemus:
Yeah, for whatever reason I've been seeing a number of intermittent build fails on 2.2.0...

geemus:
Thanks!

kbrock:
Thanks @jrafanie and @geemus

jrafanie:
@geemus Eventmachine was fixed to build against ruby 2.2 here:
https://github.com/eventmachine/eventmachine/issues/553

Upgrading to 1.0.4 of eventmachine should make this go away.

Also, here's why json is failing on 2.2:
https://github.com/flori/json/issues/229

jrafanie:
Thanks @geemus!

jrafanie:
@geemus I have the ruby 2.2 build failures recreated locally, I'll try to get a PR for your review to fix them.

geemus:
@jrafanie Awesome, much appreciated.

PR #463, opened 2015-01-27
Ruby 2.2 support was fixed in eventmachine 1.0.4.

jrafanie:
See: https://github.com/eventmachine/eventmachine/issues/553

This should fix the "undeclared identifier" build errors in travis on ruby 2.2 with eventmachine 1.0.0.

@geemus please review

PR #464, opened 2015-01-27
json 1.8.2 adds ruby 2.2 support

jrafanie:
Excon depends on json via the rdoc dependency.  We need to add a direct
dependency on json since the rdoc dependency for json (~> 1.4) doesn't enforce 1.8.2.

See: https://github.com/flori/json/issues/229

This fixes "error: too few arguments provided to function-like macro invocation" I was seeing in the travis build failures on ruby 2.2.

Between this PR and #463, I was able to bundle and run the test suite on ruby 2.2

@geemus Please review!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,8 +20,8 @@ GEM
     i18n (0.6.0)
     jruby-openssl (0.8.8)
       bouncy-castle-java (>= 1.5.0147)
-    json (1.7.3)
-    json (1.7.3-java)
+    json (1.8.2)
+    json (1.8.2-java)
     kgio (2.9.2)
     minitest (4.7.5)
     multi_json (1.3.6)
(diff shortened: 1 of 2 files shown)


geemus:
Shouldn't this have fixed the 2.2 build then?

jrafanie:
@geemus Sorry, I split up the PRs since they were separate issues so it's probably failing on the other PR's issue.

Let me change this one to build on top of #463 and so we'll get the results with both PRs.

geemus:
No worries. I suspected that might be the case, but wasn't certain from just looking at the failure cases.

jrafanie:
Ok @geemus, all :green_heart:, please review/merge #463 first, and then this PR

geemus:
Awesome, thanks for helping get to the bottom of that!

Issue #191, opened 2013-02-20
Use Addressable instead of the built-in URI


corasaurus-hex:
Addressable is a better implementation of URI in stdlib.

https://github.com/sporkmonger/addressable

I know you try to limit dependencies, and that's a good thing, but in this case it may be worth it. Or at least allow Addressable to be used in place of URI? I'd be willing to submit patches.

PR #465, opened 2015-01-27
custom uri parser to README + small fixes on Excon.new

rusikf:
1 ) Add example  to README -  ability for custom uri parser (#191)
```ruby
Excon.defaults[:uri_parser]
```
2 ) Excon as a namespace is not needed
m
3 ) bang method ```merge!``` is not needed.

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -104,7 +104,7 @@ module Excon
     #   @param [Hash<Symbol, >] params One or more option params to set on the Connection instance
     #   @return [Connection] A new Excon::Connection instance
     def new(url, params = {})
-      uri_parser = params[:uri_parser] || Excon.defaults[:uri_parser]
+      uri_parser = params[:uri_parser] || defaults[:uri_parser]
       uri = uri_parser.parse(url)
       unless uri.scheme
         raise ArgumentError.new("Invalid URI: #{uri}")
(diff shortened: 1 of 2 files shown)


lib/excon.rb
```diff
@@ -116,7 +116,7 @@ def new(url, params = {})
         :port       => uri.port,
         :query      => uri.query,
         :scheme     => uri.scheme
-      }.merge!(params)
```
geemus: I think it was intended as a means of avoiding having to allocate memory over again for this? It seems like it shouldn't hurt anything either way though. What do you think?

rusikf: I think bang method is this case is not needed
For example if we create hash
```{}``` or ```Hash.new```  and merge! it , it is not matter what we use merge! or merge because {} in this case, is not a variable(local/instance/class etc), it is freshly created object(as a "value"), which is not assigned.
*Additional:
Travic CI failed becuase json failed, not because of my pull request.*
```
An error occurred while installing json (1.7.3), and Bundler cannot continue.
```

geemus: Fair. Yeah, we fixed the json issue on master I think, so hopefully that shouldn't be an issue moving forward.

geemus:
Thanks!

Issue #471, opened 2015-02-09
readline uses read_nonblock


cydonian:
I'm using windows (8.1), cygwin (1.7.34), vagrant (1.7.2) , vagrant-openstack-plugin (0.11.1), which then uses excon 0.44.1.

Until yesterday I was using vagrant 1.6.5 and vagrant-openstack-plugin (0.9.0) installed on 24-Nov-2014. At that time excon 0.41.0 was used.

After upgrading my vagrant configuration, excon 0.44.1 was installed. Now I get an error like this

```
$ vagrant up --provider=openstack
C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/socket.rb:49:in `read_nonblock': A non-blocking socket operation could not be completed immediately. - read would block (Errno::EWOULDBLOCK) (Excon::Errors::SocketError)
        from C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/socket.rb:49:in `readline'
        from C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/response.rb:47:in `parse'
        from C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/middlewares/response_parser.rb:6:in `response_call'
        from C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/connection.rb:372:in `response'
        from C:/Users/cydonian/.vagrant.d/gems/gems/excon-0.44.1/lib/excon/connection.rb:236:in `request'
        from C:/Users/cydonian/.vagrant.d/gems/gems/fog-core-1.28.0/lib/fog/core/connection.rb:63:in `request'
        from C:/Users/cydonian/.vagrant.d/gems/gems/fog-1.27.0/lib/fog/openstack/core.rb:202:in `retrieve_tokens_v2'
        from C:/Users/cydonian/.vagrant.d/gems/gems/fog-1.27.0/lib/fog/openstack/core.rb:97:in `authenticate_v2'
        from C:/Users/cydonian/.vagrant.d/gems/gems/fog-1.27.0/lib/fog/openstack/compute.rb:395:in `authenticate'
        from C:/Users/cydonian/.vagrant.d/gems/gems/fog-1.27.0/lib/fog/openstack/compute.rb:320:in `initialize'
[...]

PR #474, opened 2015-02-11
Catch EWOULDBLOCK and IO::WaitReadable in readline

starbelly:
Fixes issue https://github.com/excon/excon/issues/471


geemus:
Thanks! Ruby's nonblocking interface is basically the worst...

PR #476, opened 2015-02-18
Don't send http request body on redirected request

Paxa:
We using squid and when POST request being redirected squid reject request (status code 411).

I find a problem in follow redirecting middleware, it should not send http body and header "Content-Length" when change to get request

### DIFF
--- a/lib/excon/middlewares/redirect_follower.rb
+++ b/lib/excon/middlewares/redirect_follower.rb
@@ -17,7 +17,11 @@ module Excon
             params = datum.dup
             params.delete(:stack)
             params.delete(:connection)
-            params[:method] = :get if [301, 302, 303].include? response[:status]
+            if [301, 302, 303].include?(response[:status])
+              params[:method] = :get
+              params.delete(:body)
+              params[:headers].delete('Content-Length')
+            end
             params[:headers] = datum[:headers].dup
             params[:headers].delete('Authorization')
             params[:headers].delete('Proxy-Connection')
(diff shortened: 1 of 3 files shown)


lib/excon/middlewares/redirect_follower.rb
```diff
@@ -17,7 +17,11 @@ def response_call(datum)
             params = datum.dup
             params.delete(:stack)
             params.delete(:connection)
-            params[:method] = :get if [301, 302, 303].include? response[:status]
+            if [301, 302, 303].include?(response[:status])
```
geemus: I guess it was historically this way also, but from rereading the spec it isn't entirely clear to me that 303 should be in this list (though 301/302 seem correct). See: http://tools.ietf.org/html/rfc7231#section-6.4

By no means is this a problem with your pull, just something worth considering/discussing while we are looking at this part of the code base.

geemus: Thoughts?

geemus:
Thanks!

lib/excon/middlewares/redirect_follower.rb
Paxa: Agree, I think it should be only 301 and 302

geemus: I'd be up for changing it. I think 301/302 is technically more correct and I'm not sure I've ever even seen a 303 in the wild.

PR #480, opened 2015-02-23
Fix reference to uninitialized variable `e`

brandur:
Refer to the exception register `$!` instead.

/cc @geemus


geemus:
released in 0.44.3

PR #485, opened 2015-03-19
Instrumentation: fix event namespace

swalkinshaw:
ActiveSupport::Notifications expects the namespace to be the suffix of the event name, not the prefix.

See [the source](https://github.com/rails/rails/blob/fcfca5c70023ab01388a87be49a4a33e29d6cd43/activesupport/lib/active_support/subscriber.rb#L67) and notice the pattern in the [Rails docs](http://edgeguides.rubyonrails.org/active_support_instrumentation.html)

Setting this properly allows the use of a `LogSubcriber` like this:

```ruby
Excon::LogSubscriber.attach_to :excon
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ connection = Excon.new(
 )
 ```
 
-Excon will then instrument each request, retry, and error.  The corresponding events are named excon.request, excon.retry, and excon.error respectively.
+Excon will then instrument each request, retry, and error.  The corresponding events are named request.excon, retry.excon, and error.excon respectively.
 
 ```ruby
 ActiveSupport::Notifications.subscribe(/excon/) do |*args|
@@ -316,7 +316,7 @@ ActiveSupport::Notifications.subscribe(/excon/) do |*args|
 end
 ```
 
-If you prefer to label each event with something other than "excon," you may specify
+If you prefer to label each event with a namespace other than "excon," you may specify
 an alternate name in the constructor:
 
 ```ruby
--- a/lib/excon/middlewares/instrumentor.rb
+++ b/lib/excon/middlewares/instrumentor.rb
@@ -3,7 +3,7 @@ module Excon
     class Instrumentor < Excon::Middleware::Base
       def error_call(datum)
         if datum.has_key?(:instrumentor)
-          datum[:instrumentor].instrument("#{datum[:instrumentor_name]}.error", :error => datum[:error])
+          datum[:instrumentor].instrument("error.#{datum[:instrumentor_name]}", :error => datum[:error])
         end
         @stack.error_call(datum)
       end
@@ -11,9 +11,9 @@ module Excon
       def request_call(datum)
         if datum.has_key?(:instrumentor)
           if datum[:retries_remaining] < datum[:retry_limit]
-            event_name = "#{datum[:instrumentor_name]}.retry"
+            event_name = "retry.#{datum[:instrumentor_name]}"
           else
-            event_name = "#{datum[:instrumentor_name]}.request"
+            event_name = "request.#{datum[:instrumentor_name]}"
           end
           datum[:instrumentor].instrument(event_name, datum) do
             @stack.request_call(datum)
@@ -25,7 +25,7 @@ module Excon
 
       def response_call(datum)
         if datum.has_key?(:instrumentor)
-          datum[:instrumentor].instrument("#{datum[:instrumentor_name]}.response", datum[:response])
+          datum[:instrumentor].instrument("response.#{datum[:instrumentor_name]}", datum[:response])
         end
         @stack.response_call(datum)
       end
--- a/tests/middlewares/instrumentation_tests.rb
+++ b/tests/middlewares/instrumentation_tests.rb
@@ -17,10 +17,10 @@ Shindo.tests('Excon instrumentation') do
 
   after do
     ActiveSupport::Notifications.unsubscribe("excon")
-    ActiveSupport::Notifications.unsubscribe("excon.request")
-    ActiveSupport::Notifications.unsubscribe("excon.response")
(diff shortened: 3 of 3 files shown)


geemus:
Did this change in Rails? I thought we were careful about how we did it at the time (some discussion/context here: https://github.com/excon/excon/pull/86). In any event, I worry that it would break things for everyone using it presently.

/cc @dpiddy @halorgium @mkb @nextmat Any thoughts/context/help?

swalkinshaw:
@geemus been like that since [at least 2010](https://github.com/rails/rails/blob/6788db824ab732b13493a9d702dd8fb89fa153c8/activesupport/lib/active_support/log_subscriber.rb#L62).

I just figured if users wanted to preserve the old functionality they just wouldn't update to the new version. 

Although I would say this is fairly important when it's supposed to support `ActiveSupport::Notifications`.

nextmat:
Hmm yeah, as I recall it has been that way for a while. I always found that super counter-intuitive, probably why we may have missed this.

mikehale:
It looks like the `Excon::LogSubscriber.attach_to :excon` usage of Notifications may be a new one for Excon. I wonder if we could build an adapter to bridge between the current internal understanding of instrumentation and what rails expects in the `attach_to` case. Putting the namespace as the suffix seems quite counter-intuitive to me as well, and I'd be hesitant to change things in excon and break existing usage.

mikehale:
This is what Excon currently supports. Notice that the names don't really matter to rails using the `instrument` + `subscribe` api.

```ruby
#!/usr/bin/env ruby

require "excon"
require "active_support"

# Subscribe to request notifications.
ActiveSupport::Notifications.subscribe(/excon.request/) do |name, started, finished, unique_id, data|
  puts name
end

# Subscribe to response notifications.
ActiveSupport::Notifications.subscribe(/excon.response/) do |name, started, finished, unique_id, data|
  puts name
end

# Register ActiveSupport::Notifications as the thing we call `instrument` on. The name passed as the 
# first paramater to insturment is opque. It simply needs to match what we expect in our subscribe calls above.

Excon.get('http://google.com', :instrumentor => ActiveSupport::Notifications)
```

```console
# $ ./instrument.rb
# excon.request
# excon.response
```

The proposed changes in this PR could be replicated with an adapter:

```ruby
#!/usr/bin/env ruby

require "excon"
require "active_support"
require "active_support/subscriber"

# Translates excon instrumentation events into rails style with
# namespace last. Yuck!
class ExconToRailsInstrumentor
  def self.instrument(name, datum, &block)
    namespace, *event = name.split(".")
    rails_name = [event, namespace].flatten.join(".")
    ActiveSupport::Notifications.instrument(rails_name, datum, &block)
  end
end

# Subscriber class that is compatible with `attach_to`.
module Excon
  class Subscriber < ActiveSupport::Subscriber
    def request(event)
      puts event.name
    end

    def response(event)
      puts event.name
    end
  end
end

# This effectively just sets up some callbacks that sort of look like this:
# * notifier.subscribe("request.excon", Proc { Excon::Subscriber.new.request(event) })
# * notifier.subscribe("response.excon", Proc { Excon::Subscriber.new.response(event) })
Excon::Subscriber.attach_to "excon"

Excon.get("https://google.com", :instrumentor => ExconToRailsInstrumentor)
```

```console
# $ ./attach_to.rb
# request.excon
# response.excon
```

My guess is that excon never considered or supported the `attach_to` api. Is that something we want to adopt? I'm not a fan of putting namespace last, but having to have an adapter for this case feels a bit weird.

@swalkinshaw is there a particular reason you reached for `attach_to` instead of `subscribe`?

swalkinshaw:
I agree that the namespace last is really weird but it's what Rails has  `¯\_(ツ)_/¯`

@mikehale I just prefer using `ActiveSupport::Subscriber` since it has a nicer interface instead of duplicating all the `subscribe` blocks. It also has the benefit of creating an `ActiveSupport::Subscriber::Event` for each.

Right now I am using the subscriptions like you have them to get around this namespace issue though. Was just hoping it would be temporary.

Are you proposing that Excon includes your adapter or just that anyone who needs it can implement it themselves?

I understand that namespace last sucks and there's the issue about backwards compatibility. But this feature was original implemented to match the ActiveSupport::Notifications "spec" which it isn't in this case. Just seems weird to me to keep it this way when any documentation/features for ActiveSupport::Notifications that specifically care about a namespace will have it last. 

That being said, it's up to you guys and if need be I'll probably just implement that adapter myself :+1:

mikehale:
@swalkinshaw does this syntax make the current state of things any more palatable?

```ruby
ActiveSupport::Notifications.subscribe(/excon/) do |*args|
  event = ActiveSupport::Notifications::Event.new(*args)
  puts event.name
end
```

swalkinshaw:
@mikehale that's actually exactly what I'm currently doing.

However that still doesn't address the incompatibility with `ActiveSupport::Notifications` as a whole. If the namespace won't be changed I think it would at be least useful to include a note in the `README` about this discrepancy.

mikehale:
@swalkinshaw gotcha. I'm not sure that changing the current default behavior of excon is the right thing to do, but a note in the readme sounds like a good next step. Do you mind taking a stab at a PR with the readme change?

geemus:
I'm amazed we are just realizing the discrepancy, but I guess the namespace first version feels more natural (to me at least), so maybe it was just all too easy to overlook. Seems like a note in the README would be a great place to start. Maybe making it either more configurable or having a parallel implementation that swapped them can follow, but starting simple/small seems like the right first step I think. Thanks!

swalkinshaw:
README updates: https://github.com/excon/excon/pull/489

Up to you which one to merge in :)

geemus:
Cool, I merged the other one for now, but we should definitely keep an eye on it over time.

PR #487, opened 2015-03-23
Use the proper IO.select with a connect timeout for SSL sockets

zanker:
I don't have a good test case that lets someone else replicate this. One of our apps is getting SSL_connect SYSCALL returned=5 errno=0 state=SSLv3 read server hello A. Net::HTTP works fine with the same certs, another app works fine with the same OpenSSL/Ruby/Excon version.

For some reason one app gets those errors with Excon, I narrowed it down to the fact that Excon is repeatedly calling connect_nonblock and using a raise to "block". As soon as I switch it to the proper method of using IO.select it works fine.

Given that connect_nonblock can be called 10s/100s of times, that makes this raise quite a lot which is really inefficient, especially on JRuby.

Also I can't see a reason to do it that way at this point, because non-SSL uses `IO.select` anyway.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -116,20 +116,17 @@ module Excon
 
       begin
         if @nonblock
-          loop do
-            begin
-              @socket.connect_nonblock
-              break # connect succeeded
-            rescue OpenSSL::SSL::SSLError => error
-              # would block, rescue and retry as select is non-helpful
-              raise error unless error.message == 'read would block'
-            end
+          begin
+            @socket.connect_nonblock
+          rescue IO::WaitReadable
+            IO.select([@socket])
+            retry
           end
         else
           @socket.connect
         end
-      rescue OpenSSL::SSL::SSLError => e
-        raise e
+      rescue OpenSSL::SSL::SSLError
+        raise
       rescue
         raise Excon::Errors::Timeout.new('connect timeout reached')
       end


geemus:
Sounds good, unfortunately it looks like this is breaking a couple tests. Could you take a look and see if you can figure out where the discrepancy might be? Thanks!

zanker:
Probably because I changed the exception rescues a bit, but I reverted that and will see if that fixes it.

zanker:
There we go, green now. Sorry for the delay

lib/excon/ssl_socket.rb
```diff
@@ -116,20 +116,17 @@ def initialize(data = {})
...
-              @socket.connect_nonblock
-              break # connect succeeded
-            rescue OpenSSL::SSL::SSLError => error
-              # would block, rescue and retry as select is non-helpful
-              raise error unless error.message == 'read would block'
-            end
+          begin
+            @socket.connect_nonblock
+          rescue IO::WaitReadable
+            IO.select([@socket])
```
jcsalem: I think you should select on socket errors as well:
IO.select([@socket],nil,[@socket])

Also, don't you need to select on @socket.io rather than just @socket?

zanker: `IO.select` automatically calls `to_io`. 

If you select on socket errors, wouldn't that mean you retry forever effectively?

jcsalem: You can ignore my comment on the 3rd argument to IO.select.  Even thought it's called **exceptfs**, it's for exceptional conditions not exceptions/errors as I've always assumed.  This man page has more but it's basically only used for out-of-band data which doesn't apply in this case: http://man7.org/linux/man-pages/man2/select_tut.2.html

Given that, your patch looks good to me.

geemus:
Thanks!

geemus:
Fix out in 0.45.0

jcsalem:
Thanks for the quick turnaround!

PR #488, opened 2015-03-24
Use our default config SSL config then ENV when available

zanker:
This changes Excon to allow overriding SSL certs based on the config in the connection instead of defaulting to the ENV variable.

Google's [signet](https://github.com/google/signet/blob/master/lib/signet/ssl_config.rb) gem automatically sets these ENV variables, and it leads to unintuitive failures when you're using S2S and the signet gem overrides your config.


geemus:
Seems reasonable I think/hope. Thanks!

geemus:
Fix out in 0.45.0

PR #489, opened 2015-03-25
Update ActiveSupport::Notifications README section

swalkinshaw:
Documents the incompatibility with Rails' default event name format.

See https://github.com/excon/excon/pull/485 for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ connection = Excon.new(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


geemus:
Awesome, thanks. I think starting with this kind of "simplest that could work" doc fix is great. Definitely open to continuing to discuss and/or revisit in the future, but glad to have this for now.

geemus:
Thanks!

Issue #467, opened 2015-01-29
Bad address error with openssl 1.0.2


matkam:
I'm not sure if this is a bug with Ruby OpenSSL, excon, or possibly even Fog. Running Mac OS X with Ruby 2.2.0 or 2.1.3 installed via ruby-build and openssl 1.0.2 installed via Homebrew. I get the following error when uploading file more than ~1MB (maybe even less) to AWS S3:
```
/Users/matt/.rbenv/versions/2.2.0/lib/ruby/2.2.0/openssl/buffering.rb:383:in `syswrite_nonblock': Bad address (Errno::EFAULT) (Excon::Errors::SocketError)
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/2.2.0/openssl/buffering.rb:383:in `write_nonblock'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/socket.rb:109:in `write'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/connection.rb:151:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/middlewares/mock.rb:47:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/middlewares/instrumentor.rb:22:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/middlewares/base.rb:15:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/middlewares/base.rb:15:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/middlewares/base.rb:15:in `request_call'
	from /Users/matt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/excon-0.43.0/lib/excon/connection.rb:232:in `request'
[...]

PR #490, opened 2015-03-25
Catches issue with 1.0.2 OpenSSL which leads to a fault.

imajes:
See issue #467 for more info.

@geemus - this is a prospective fix, happy to take a different approach if you have thoughts.

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -241,6 +241,14 @@ module Excon
           # I wish that this API accepted a start position, then we wouldn't
           # have to slice data when there is a short write.
           written = @socket.write_nonblock(data)
+        rescue Errno::EFAULT
+          if OpenSSL::OPENSSL_LIBRARY_VERSION.split(' ')[1] == '1.0.2'
+            msg = "The version of OpenSSL this ruby is built against (1.0.2) has a vulnerability
+                   which causes a fault. For more, see https://github.com/excon/excon/issues/467"
+            raise SecurityError.new(msg)
+          else
+            raise error
+          end
         rescue OpenSSL::SSL::SSLError, Errno::EAGAIN, Errno::EWOULDBLOCK, IO::WaitWritable => error
           if error.is_a?(OpenSSL::SSL::SSLError) && error.message != 'write would block'
             raise error


lib/excon/socket.rb
```diff
@@ -241,6 +241,14 @@ def write_nonblock(data)
           # I wish that this API accepted a start position, then we wouldn't
           # have to slice data when there is a short write.
           written = @socket.write_nonblock(data)
+        rescue Errno::EFAULT
+          if OpenSSL::OPENSSL_LIBRARY_VERSION.split(' ')[1] == '1.0.2'
+            msg = "The version of OpenSSL this ruby is built against (1.0.2) has a vulnerability
+                   which causes a fault. For more, see https://github.com/excon/excon/issues/467"
+            raise SecurityError.new(msg)
```
geemus: I hadn't seen `SecurityError` before. I can find it in new Ruby docs, but how far back is it implemented? I just don't want to have a backwards incompatible thing, so StandardError or something might be safer (unless we know for sure SecurityError has been around for a while, definitely possible I just haven't run across it before).

geemus:
Approach looks great, just some small concern around backwards compatibility if you could take a look.

lib/excon/socket.rb
imajes: Yeah, i just looked up ruby standard exceptions and that one stood out. will do a quick bit of research to see how far back it goes. Excon supports 1.8.7+ ?

geemus: Yeah, for now at least we are still trying to support 1.8.7+. I imagine we'll start dropping old versions at some point, but we haven't really had particularly strong reasons to start down that path so far.

imajes:
@geemus: seems ```SecurityError``` exists in 1.8.7: http://ruby-doc.org/core-1.8.7/SecurityError.html - sound good to go then?

imajes:
@geemus: ping- just checking in on this to see if we can wrap it up?

geemus:
Thanks for double checking, yeah seems great (just a little swamped today).

geemus:
Fix out in 0.45.0

PR #491, opened 2015-03-25
with_defaults method — allows defaults to be overridden temporarily within a block

kch:
@geemus per discussion in https://github.com/interagent/heroics/pull/62

no tests because I didn't find any tests related to defaults.

I also froze the `DEFAULTS` constant because having a mutable constant seemed like some state corruption waiting to happen.

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -44,7 +44,8 @@ module Excon
     def defaults
       defaults = Thread.current.thread_variable_get(:excon_defaults)
       return defaults if defaults
-      Thread.current.thread_variable_set(:excon_defaults, DEFAULTS.dup)
+      defaults = (Thread.main.thread_variable_get(:excon_defaults) || DEFAULTS).dup
+      Thread.current.thread_variable_set(:excon_defaults, defaults)
     end
 
     # Change defaults for Excon connections
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -42,13 +42,23 @@ module Excon
 
     # @return [Hash] defaults for Excon connections
     def defaults
-      @defaults ||= DEFAULTS
+      defaults = Thread.current.thread_variable_get(:excon_defaults)
+      return defaults if defaults
+      Thread.current.thread_variable_set(:excon_defaults, DEFAULTS.dup)
     end
 
     # Change defaults for Excon connections
     # @return [Hash] defaults for Excon connections
     def defaults=(new_defaults)
-      @defaults = new_defaults
+      Thread.current.thread_variable_set(:excon_defaults, new_defaults)
+    end
+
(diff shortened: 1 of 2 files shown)


lib/excon.rb
```diff
@@ -42,13 +42,23 @@ class << self
 
     # @return [Hash] defaults for Excon connections
     def defaults
-      @defaults ||= DEFAULTS
+      defaults = Thread.current.thread_variable_get(:excon_defaults)
```
kch: wonder if the variable name should be more complex, since the namespace is (thread-)global. ideally ruby would have gensyms, but nope.

kch:
There may be a problem if you update defaults on the main thread. We'll still be copying the original `DEFAULTS` constant on new threads. So maybe back to mutable constant behaviour, plus mutexes.

lib/excon.rb
```diff
@@ -42,13 +42,23 @@ class << self
...
+      return defaults if defaults
+      Thread.current.thread_variable_set(:excon_defaults, DEFAULTS.dup)
     end
 
     # Change defaults for Excon connections
     # @return [Hash] defaults for Excon connections
     def defaults=(new_defaults)
-      @defaults = new_defaults
+      Thread.current.thread_variable_set(:excon_defaults, new_defaults)
```
kch: I kinda would prefer to do `new_defaults.dup` here. Values in it can be mutated later, and currently that would mutate the hash in the caller too.

kch:
FWIW I don't want this to be merged as-is, without further discussion. @geemus are we interested in pursuing this idea further? If not close, otherwise find me in hipchat?

geemus:
Hmm. It is an interesting line of inquiry, and one for which I don't have a ready answer. Maybe we could do a global default which would merge in or override the thread-local ones or something. Definitely starts to make me wonder if the complexity involved is really worth it. The defaults were meant to just simplify particular use cases (and I think they do), but I hadn't considered that it mostly simplifies only in the non-concurrent cases. Happy to discuss some more, but a bit swamped. Let me know if this is (or becomes) a priority for you and I'll be sure to carve out more time to think about and discuss it. Thanks!

kch:
The thread stuff is just icing on the cake.

The main issue is that I often find myself using libraries that are implemented using excon but completely opaque in terms of providing a way to change excon specifics (say, heroics). Yes global defaults help, but different requests need different options.

So a way to wrap any call that eventually makes excon requests with a particular set of defaults is the main focus here.

geemus:
Yeah, that makes sense. I'll try to think about this and work on something
soon (bit swamped presently).

geemus:
Closing, as there didn't seem to be too much demand and it would increase complexity.

PR #492, opened 2015-03-25
Fix CPU spin wait on SSL connections

jcsalem:
We were seeing excessive CPU consumption when initiating SSL connections on nonblocking sockets via Excon. When Excon creates an SSL connection on a non-blocking socket, it currently retries the connect call in a tight loop until the connection completes. This needlessly uses CPU cycles.

This patch inserts a IO.select call before it retries the connect thereby allowing the CPU to do useful work and only waking up the calling process when the connection completes (or errors).

This has been tested under heavy production workload and significantly reduces CPU consumption while working reliably.

NOTE: One could argue that the call should return a "read would block" error to Excon's caller, however, that would be a significant functionality change.  This patch keeps the behavior identical to current functionality but removes the CPU spin wait.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -121,8 +121,9 @@ module Excon
               @socket.connect_nonblock
               break # connect succeeded
             rescue OpenSSL::SSL::SSLError => error
-              # would block, rescue and retry as select is non-helpful
               raise error unless error.message == 'read would block'
+              # Wait for socket status to change
+              IO.select([@socket.io],nil,[@socket.io])
             end
           end
         else


zanker:
Hah we found the same issue around the same time. https://github.com/excon/excon/pull/487 you can rescue a more specific exception to avoid needing to check content.

jcsalem:
I agree that #487 will fix this (though I have one comment on that PR I'd like addressed.) 
I'll close this PR if #487 is merged.

geemus:
Hey, just merged #487 (and I think the comment got resolved there if I'm not mistaken). Thanks!

PR #494, opened 2015-03-27
Fix scope of buffer in readline

starbelly:
Because buffer was declared within the scope of the begin/end block in readline(),  data read in initially would become lost to oblivion if a false-positive exception was raised and the block re-tried. This would result in lines like  "TTP/1.1 400 Bad Request\r\n", where H was at the start of the buffer, an exception was raised, and the block re-tried if it was a false positive, thus loosing the 'H'

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -40,8 +40,8 @@ module Excon
 
     def readline
       return legacy_readline if RUBY_VERSION.to_f <= 1.8_7
+      buffer = ''
       begin
-        buffer = ''
         buffer << @socket.read_nonblock(1) while buffer[-1] != "\n"
         buffer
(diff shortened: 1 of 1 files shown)


tokengeek:
This looks good.

The header parsing issue doesn't appear using this patch even making 100 calls. Without it, error raises within a few attempts.

johnl:
Any thoughts how to catch problems like this future? We caught it because we were running acceptance tests where excon was involved (it manifested as lots of weird random socket errors).

Does excon test suite have some remote tests?

Sorry if this isn't the right place to discuss this!

geemus:
@starbelly awesome, thanks for looking. I imagine I would have had much less luck fixing that so quick.

@tokengeek thanks for verifying.

@johnl hmm. We do run servers/test them (but all locally). I'm not sure how we would/could force it into the situation where this was occurring. Open to suggestions though.

PR #496, opened 2015-03-27
Complete response test

starbelly:
Added unit tests to make sure we're not missing in chars in our response data


geemus:
@starbelly Only managed to glance at this so far, hope to dig in tomorrow. Thanks!

geemus:
Cool, I'll bring this in, but I think maybe I'm going to rename `raw_status` to just `status_line` to be explicit. Thanks again for your help around this!

PR #498, opened 2015-03-31
Catch IO::WaitWritable for SSL sockets too

zanker:
Most likely this is overkill. The original SSL code in Excon didn't catch this, but I found out Ruby docs do recommend it so being safe: http://ruby-doc.org/stdlib-1.9.3/libdoc/openssl/rdoc/OpenSSL/SSL/SSLSocket.html


geemus:
Seems good, better safe than sorry (and it shouldn't really cost us anything either way).

PR #501, opened 2015-04-13
Allow connect errors to bubble up

### LINKED ISSUE #500: SSLSocket "connect timeout reached" error may mask other errors
I'm worried that the [connection code](https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb#L117-L135), maybe be masking other types of errors with a timeout error.

```ruby
      begin
        if @nonblock
          begin
            @socket.connect_nonblock
          rescue IO::WaitReadable
            IO.select([@socket])
            retry
          rescue IO::WaitWritable
            IO.select(nil, [@socket])
            retry
          end
        else
          @socket.connect
        end
      rescue OpenSSL::SSL::SSLError
        raise
      rescue
        raise Excon::Errors::Timeout.new('connect timeout reached')
      end
```
Is there something we could do here instead of always raising `Excon::Errors::Timeout`?

mikehale:
Fixes #500

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -114,24 +114,18 @@ module Excon
         @socket.hostname = @data[:host]
       end
 
-      begin
-        if @nonblock
-          begin
-            @socket.connect_nonblock
-          rescue IO::WaitReadable
-            IO.select([@socket])
-            retry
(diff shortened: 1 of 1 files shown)


geemus:
Leaning toward fix on #500 instead.

PR #500, opened 2015-04-14
SSLSocket "connect timeout reached" error may mask other errors

geemus:
I'm worried that the [connection code](https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb#L117-L135), may be masking other types of errors with a timeout error.

```ruby
      begin
        if @nonblock
          begin
            @socket.connect_nonblock
          rescue IO::WaitReadable
            IO.select([@socket])
            retry
          rescue IO::WaitWritable
            IO.select(nil, [@socket])
            retry
          end
        else
          @socket.connect
        end
      rescue OpenSSL::SSL::SSLError
        raise
      rescue
        raise Excon::Errors::Timeout.new('connect timeout reached')
      end
```
Is there something we could do here instead of always raising `Excon::Errors::Timeout`?

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -130,7 +130,7 @@ module Excon
         end
       rescue OpenSSL::SSL::SSLError
         raise
-      rescue
+      rescue Errno::ETIMEDOUT
         raise Excon::Errors::Timeout.new('connect timeout reached')
       end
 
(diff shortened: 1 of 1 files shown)


geemus:
Commits added, converted to pull request.

mikehale:
:+1:

geemus:
Error appears to be weird and unrelated, bringing this in.

mikehale:
This has been helpful in diagnosing issues for me. Instead of seeing `Excon::Errors::Timeout`, I now see things like `Connection reset by peer - SSL_connect (Errno::ECONNRESET)`.

geemus:
Awesome, thanks for the update.

geemus:
Released in 0.45.2

Issue #505, opened 2015-04-30
Undefined local variable error


KellyMahan:
Error isn't defined at this location. For some reason when using large images with dragonfly it errors out here.

in dragonfly with s3 storage thumb("1200x1200>") works just fine
but thumb("1600x1600>") throws the error below

https://github.com/excon/excon/commit/8f741167125bb6fe744e534d1d80c358ca0de657

KellyMahan:
pull request https://github.com/excon/excon/pull/506

KellyMahan:
After referencing the gem from my repository with the "fix" this is the stack trace i get.

Excon::Errors::SocketError - Bad address (Errno::EFAULT):
  .rbenv/versions/2.1.2/lib/ruby/2.1.0/openssl/buffering.rb:383:in `syswrite_nonblock'
  .rbenv/versions/2.1.2/lib/ruby/2.1.0/openssl/buffering.rb:383:in `write_nonblock'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/socket.rb:243:in `block in write_nonblock'
   () /.gems/bundler/gems/excon-71007862fcc4/lib/excon/socket.rb:238:in `loop'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/socket.rb:238:in `write_nonblock'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/socket.rb:78:in `write'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/connection.rb:152:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/mock.rb:47:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/instrumentor.rb:22:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/base.rb:15:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/base.rb:15:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/base.rb:15:in `request_call'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/connection.rb:233:in `request'
   () .gems/bundler/gems/excon-71007862fcc4/lib/excon/middlewares/idempotent.rb:26:in `error_call'
[...]

PR #506, opened 2015-04-30
was this the original intent?

KellyMahan:
line 250 raise error was undefined

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -241,7 +241,7 @@ module Excon
           # I wish that this API accepted a start position, then we wouldn't
           # have to slice data when there is a short write.
           written = @socket.write_nonblock(data)
-        rescue Errno::EFAULT
+        rescue Errno::EFAULT => error
           if OpenSSL.const_defined?(:OPENSSL_LIBRARY_VERSION) && OpenSSL::OPENSSL_LIBRARY_VERSION.split(' ')[1] == '1.0.2'
             msg = "The version of OpenSSL this ruby is built against (1.0.2) has a vulnerability
                    which causes a fault. For more, see https://github.com/excon/excon/issues/467"


imajes:
@KellyMahan that's a really interesting catch. Honestly, it shouldn't have gotten into that branch -- and i didn't test for that. 

but, on a larger note, it's really interesting as to why you'd hit that *at all*. what did the error throw when you changed that line?

KellyMahan:
third comment on https://github.com/excon/excon/issues/505

imajes:
@KellyMahan ok, that's weird, that comment didn't show for me when i loaded it first...

Can you share your OpenSSL lib, binary, and client versions?

KellyMahan:
osx with latest updates

$ which openssl
/usr/bin/openssl

$ openssl version
OpenSSL 0.9.8zd 8 Jan 2015

$ brew list openssl
/usr/local/Cellar/openssl/1.0.2/bin/openssl

not sure which one is actually in use

imajes:
@KellyMahan: ``` puts OpenSSL::OPENSSL_VERSION```

KellyMahan:
OpenSSL 1.0.1i 6 Aug 2014

imajes:
@KellyMahan: that's vulnerable and out of date. You need to recompile ruby against the newer openssl (that which is in your brew cellar should be fine, if you recently ran brew update/upgrade). You'll want to run something like this, assuming you use ruby-build:

```RUBY_CONFIGURE_OPTS=--with-openssl-dir=/usr/local/Cellar/openssl/1.0.2 rbenv install ...```

imajes:
right now mine reports as ```OpenSSL 1.0.2a 19 Mar 2015```, which has a patch to fix the non blocking IO problem.

KellyMahan:
upgrading openssl fixed the bug dragonfly. Thanks

geemus:
Thanks! Also glad you got your setup fixed, but we'll also bring this in for completeness.

PR #508, opened 2015-05-11
Add option to retain cookies through redirects

colindean:
I encountered a use case that necessitated passing cookies provided by the first URL to redirect URL. 

I've disabled this functionality by default in this PR.

See the commit messages for a little more history and discussion.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -12,6 +12,7 @@
 * Carl Hörberg <carl.hoerberg@gmail.com>
 * Carlos Sanchez <csanchez@maestrodev.com>
 * Claudio Poli <masterkain@gmail.com>
+* Colin Dean <colindean@us.ibm.com>
 * Damien Mathieu <damien@heroku.com>
 * Dan Hensgen <dan@methodhead.com>
 * Dan Peterson <dpiddy@gmail.com>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,6 +9,7 @@ GEM
     activesupport (3.2.6)
       i18n (~> 0.6)
       multi_json (~> 1.0)
+    backports (3.6.4)
     bouncy-castle-java (1.5.0147)
     chronic (0.6.7)
     delorean (2.0.0)
@@ -29,6 +30,8 @@ GEM
     rack (1.6.0)
     rack-protection (1.2.0)
       rack
+    rack-test (0.6.3)
+      rack (>= 1.0)
     raindrops (0.13.0)
     rake (0.9.2.2)
     rdoc (3.12)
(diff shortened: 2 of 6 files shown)


colindean:
It'd be awesome if this could make it into a release before 5 June 2015.

geemus:
I'll try to look more deeply at this soon (getting married in a couple days, so a bit distracted). You should be able to run your own middleware to provide this in the mean time if needed in the mean time though. Is there a particular reason for the timeline you suggested?

colindean:
We're using our own middleware for now, but we'd love to ship our product using an upstream release instead of our internal branch and build. 5 June would be great, but any time before 12 June would be OK. After that and we'll have to ship with the internal. It's not the end of the world; please don't let this get in the way of wedding plans. Congratulations!

geemus:
Sure thing. I can probably look a bit more closely next week. One thing I wonder about though, it seems like this isn't wholly related to just redirecting (ie you might want to set this for subsequent requests also I would imagine). What do you think about making this it's own middleware separate from redirect? As long as it is earlier in the stack it should still have the behavior you want, but it would keep the logic in both more simple (and allow for this to be used independently if that was desirable). Does that make sense? What do you think?

colindean:
A separate middleware might not be a bad idea for a future iteration.  I think this patch is sufficient as-is. 

I didn't investigate if any other parts kept state when I did this patch.

geemus:
I guess I had imagined it would push data in to datum, but I suppose it would need to push all the way out to the connection settings if it was to be a persistent value (so merging it in to headers within @data on the connection). Does that make sense? I think I would prefer to keep this in it's own distinct middleware for now if possible. If nothing else, it would be easy to add it to the redirect middleware later (whereas it would be pretty hard to move it out without breaking things for people that might be relying on it). Does that make sense? Would you be up for attempting the change in that style? Thanks!

colindean:
That could work, but I don't know when is have a chance to do it. Maybe in a couple of weeks.

djberg96:
Seems reasonable to me. @colindean Looks like you'll need to rebase.

colindean:
Ugh, email responses didn't go through…

----

I no longer am set up to be able to do so without a good bit of time investment. I no longer have the OS installation on which I did the work and am no longer employed by the company for which I did the work.

I'll add it to my list of things to do but I might not get around to it for a long time. It would be really awesome if someone else could pick this up and run with it. @atambo still works there and is probably responsible for the internal fork of excon that includes this patchset. He might be equipped to handle this sooner with the incentive to longer maintain that internal fork ;-)

geemus:
@colindean ah, got it. Thanks for the detailed update.

Issue #360, opened 2014-01-13
Excon::Utils#query_string should convert array parameter properly


omikronn:
Say, I need to do some request to some API, where some query parameters look like `param[]=value1&param[]=value2`. If I send a hash of the form `{ param: [value1, value2] }`, `Excon::Utils#query_string` will convert this to `&param=value1&param=value2` where effectively only `value2` will be taken into consideration.

I have a tested patch for this (really, just a couple of lines of code).
What do you guys think? Would this make a nice addition?

sodabrew:
Would you be amenable to importing this module from HTTParty?
https://github.com/jnunemaker/httparty/blob/master/lib/httparty/hash_conversions.rb
http://www.rubydoc.info/github/jnunemaker/httparty/HTTParty/HashConversions.to_params

geemus:
Hmm. I think the bigger question we had above wasn't how to get it working (it could be done in the current codebase with maybe 2 lines changed), but more if that should be the default behavior and/or if it should be configurable/toggleable. Unfortunately, I suspect it would take a good deal more code (and complexity) to support the more flexible version of this.

Perhaps we should just do it this way though, it seems a quite common pattern, but I wouldn't say it is universal. Similarly as an end user, I think you can already get this behavior by just naming the key that you know has an array as `key[]`, if I'm not mistaken. So in that sense it should already be workable, assuming it is straightforward to know which values will be arrays. Does that help/make sense?

sodabrew:
The module from HTTParty also encodes `{:foo => { :bar => 'val', :qux => 'val' } }` as `foo[bar]=val&foo[qux]=val`, which is typical of HTTP form encoding. Let me try out `{ 'foo[]' => ['val1', 'val2'] }` though. I'll look into a PR to add tests for these to https://github.com/excon/excon/blob/master/tests/query_string_tests.rb

geemus:
Yeah. I think my hesitation is just that it is a de facto standard (not from an RFC), so it may be common, but I don't think it is official and I'm not at all certain it is universal. So forcing it as the default might be bad (and is probably a breaking change). If specifying it explicitly works as we suspect and we added tests, it seems like that might be a nice compromise for now anyway. Thanks!

sodabrew:
PR for array elements. Hash elements are not handled at all, so that's certainly a deficiency of the current code. Included a test for such mangling.

geemus:
True, I guess I hadn't run into nested hashes as much.

PR #516, opened 2015-06-26
Test that the query string can serialize an array as repeated query elements

sodabrew:
Following up from #360, this test shows that an array is serialized as repeated query elements. To get the Rails / PHP array behavior, set the key to 'foo[]' => [ array elements ] to product the query string `?foo[]=bar&foo[]=baz`.

### DIFF
--- a/tests/query_string_tests.rb
+++ b/tests/query_string_tests.rb
@@ -1,57 +1,68 @@
 Shindo.tests('Excon query string variants') do
   with_rackup('query_string.ru') do
-    connection = nil
+    connection = Excon.new('http://127.0.0.1:9292')
 
     tests(":query => {:foo => 'bar'}") do
+      response = connection.request(:method => :get, :path => '/query', :query => {:foo => 'bar'})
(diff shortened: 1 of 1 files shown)


tests/query_string_tests.rb
```diff
@@ -1,57 +1,67 @@
...
         query_string.split('&').include?('foo=bar')
       end
 
       test("query string sent includes 'me=too'") do
         query_string.split('&').include?('me=too')
       end
+    end
+
+    # To emulate the Rails and PHP style of serializing an array, use { 'foo[]' => [ array elements ] }
+    tests(":query => {:foo => ['bar', 'baz'], :me => 'too'}") do
```
geemus: Maybe, for full clarity we could just have the test use the `[]` style explicitly? That way it would be clearer the purpose for this test. What do you think?

sodabrew: Sure, I actually did that first but switched it before I pushed. Will update the commit in a moment.

geemus:
Awesome, thanks!

PR #518, opened 2015-07-21
Fix Bad Argument on SocketError Initializer

swithrow:
Was `Excon::Error` (bad constant), replaced with `Excon::Errors::Error`.

Otherwise `Excon::Error::SocketError.new` dies with undefined constant `Excon::Error`


geemus:
Thanks!

Issue #519, opened 2015-07-22
Unusual encoding for a feed


mahemoff:
Any idea why this feed is processed differently by excon compared to open-uri?

> url = http://insertmoin.de/podcast/

> open(url).read[0,100]
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>        <rss version=
\"2.0\"\n            xmlns:content=\"http://p"

> Excon.new(url).request(method: 'GET').body[0,100]
"\x1F\x8B\b\x00\x00\x00\x00\x00\x00\x03\xED\xBD\xDBr\xDBH\x96.|=
\xFB)\xB2\xD8\x11\x1D\xAE\x18%\xC5\x93N\xAC\xEA\xAA_\x96d\xC9e\x
CB\xD6\x88\xB2\xDDS;v0\x92D\x92@\x11\av& Z\xBA\x9A\x8B\xFD\x18\x
FD\xFF\xA5\x1D\xE1\x9B\xFD\x02\xBE\xF2\x95\xF8&\xF3$\xFFZ+\x01\x
1E$\xD1\x96@\xCA-\x90\xA8\x19\xB7$\x12\xC8Ld\xAE"

geemus:
My guess would be that it is gzip or compress encoded. By default excon gives back raw responses, though you can mount the decompress middleware to expand/unzip it. You can probably confirm by checking the encoding value in the headers though.

mahemoff:
That did it, thanks. I'm just surprised it didn't come up before.

I went with this:

    case res.status
      when 200, 304
        body =
          if res[:headers]['Content-Encoding']=='gzip'
            Zlib::GzipReader.new(StringIO.new(res.body)).read
          else
            res.body
          end
      when 404
        ...

geemus:
Cool, yeah, I would expect it more often also. Usually most servers will only give you back gzip/compress if you ask for it though (which you certainly can and might even want to), but this server maybe always gives it back, even though it technically shouldn't do so. It's always a bit of a tradeoff depending on size of request and type of request. ie it is less data over the wire, but how much less depends on the type of data and how compressable it is. Plus, though it might then be faster to transfer, you have to pay the cost in time of encoding/decoding it (which in small requests or things that don't compress well, can mean it is slower than if you didn't. Anyway, glad this is now sorted out for you.

mahemoff:
[...]

PR #520, opened 2015-07-24
README shows options that are sometimes default

mahemoff:
Sometimes Excon uses different default options than other clients, so highlighting them might help with debugging, as I've sometimes submitted issues comparing excon to curl or open-uri.

(cf #519)


geemus:
Thanks!

PR #522, opened 2015-08-04
Allows SSL verification disabling via ENV var

### LINKED ISSUE #521: Possible to disable SSL verification via ENV variable
When encountering an SSL certificate verification error, the error is "Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure)."

The final option disables SSL verifcation entirely and solves my problem, but there is no way to enable this via an environment variable like the other options.  Is this something that could be added (`EXCON_SSL_VERIFY=false` or something)? Happy to do the work myself if I know it's something that will be accepted.

wadetandy:
Fixes #521.  Previously was possible to set certificates via environment
variables, but not possible to disable verification entirely.

I haven't used Shindo for writing tests before so glad to re-organize if there's a better way for these tests to be done.

### DIFF
--- a/lib/excon/errors.rb
+++ b/lib/excon/errors.rb
@@ -10,7 +10,7 @@ module Excon
 
       def initialize(socket_error=Excon::Errors::Error.new)
         if socket_error.message =~ /certificate verify failed/
-          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure).")
+          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` or `ENV['EXCON_SSL_VERIFY_PEER'] = 'false'` (less secure).")
         else
           super("#{socket_error.message} (#{socket_error.class})")
         end
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -4,6 +4,8 @@ module Excon
       OpenSSL::SSL::SSLSocket.public_method_defined?(m)
     end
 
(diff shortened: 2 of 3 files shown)


lib/excon/errors.rb
```diff
@@ -10,7 +10,7 @@ class SocketError < Error
 
       def initialize(socket_error=Excon::Errors::Error.new)
         if socket_error.message =~ /certificate verify failed/
-          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure).")
+          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` or `ENV['EXCON_SSL_VERIFY_PEER'] = 'false'` (less secure).")
```
geemus: Maybe we should just have this be `SSL_VERIFY_PEER`, I don't think it should conflict, and none of the other options here have the `EXCON` prefix, what do you think?

lib/excon/ssl_socket.rb
```diff
@@ -4,6 +4,8 @@ class SSLSocket < Socket
       OpenSSL::SSL::SSLSocket.public_method_defined?(m)
     end
 
+    VERIFY_ENV_VAR = 'EXCON_SSL_VERIFY_PEER'
```
geemus: Maybe just for clarity/explicitness this could be `VERIFY_PEER_ENV_VAR` for the constant. We might also want to consider putting this into `lib/excon/constants` as that is where most all of the others live.

lib/excon/ssl_socket.rb
```diff
@@ -26,7 +28,9 @@ def initialize(data = {})
         ssl_context.ssl_version = @data[:ssl_version]
       end
 
-      if @data[:ssl_verify_peer]
+      env_verify_peer = ENV.has_key?(VERIFY_ENV_VAR) && (ENV[VERIFY_ENV_VAR] && ENV[VERIFY_ENV_VAR] != 'false')
```
geemus: It took me a couple reads to parse this, though it probably should have been easier for me. Maybe we could do `!([nil,'false'].include?(ENV[VERIFY_ENV_VAR])` to be a little more explicit? What do you think?

geemus:
Closing due to inactivity and merge conflicts, please let us know if you would like to revisit.

PR #523, opened 2015-08-11
give the instrumentor all the information

lanej:


lanej:
@geemus may warrant some discussion because it changes instrumentor behavior, but it is very useful to have the entire request-response structure available.

lib/excon/constants.rb
```diff
@@ -119,10 +119,10 @@ module WaitWritable; end
     :idempotent           => false,
     :instrumentor_name    => 'excon',
     :middlewares          => [
+      Excon::Middleware::Instrumentor,
```
lanej: having it above `Idempotent` means you don't log the retry requests, but having it below `Idempotent` means you don't have access to the parsed response.

geemus: Yeah, I think we have run into this before. Not sure what the best approach would be.

geemus:
I wonder if it would make sense for this to live as it's own middleware (so as to avoid needing to change the order of other things). It does seem strange though, as other things like this live together within instrumentor middleware.

lanej:
@geemus that seems like a decent idea.  I could write a different instrumentor and just not insert it into the default middleware by default.  Maybe just doc it and where to insert it into the stack ?

geemus:
Yeah, that seems like a good place to start at least. That will get it to a place that is easily/nicely accessible without having to worry as much about backwards compatibility and stuff. Then, as we get a bit more experience/comfort we can revisit (ie maybe we could rearrange or absorb it into intsrumentation middleware prior to 1.0 or something).

lanej:
@geemus seems wise. I'll give that a shot.

geemus:
Great, thanks!

lanej:
a slightly more coherent version

```ruby
class ExconInstrumentor < Excon::Middleware::Base
  def error_call(datum)
    ActiveSupport::Notifications.instrument("excon.error", datum) {
      @stack.error_call(datum)
    }
  end

  def request_call(datum)
    event_name = if datum[:retries_remaining] < datum[:retry_limit]
                   "excon.retry"
                 else
                   "excon.request"
                 end
    ActiveSupport::Notifications.instrument(event_name, datum) do
      @stack.request_call(datum)
    end
  end

  def response_call(datum)
    ActiveSupport::Notifications.instrument("excon.response", datum) {
      @stack.response_call(datum)
    }
  end
end
```

obviously wouldn't use `ActiveSupport::Notifications` directly but this works in such a way that the complete set of information is given to the subscriber.

lib/excon/constants.rb
geemus: Makes me wonder if there should be a retry callback also (or something to that effect). The current setup somewhat limits what can be done here.

Perhaps the instrumentor could check for retries and behave differently? It should be able to figure out via comparing to retries_remaining. Though it may be better/cleaner to actually add a retry boolean or something to the data, or at least a method that did the comparison instead of leaking that out across even more places.

geemus: Additionally, this perhaps lends additional credence to: https://github.com/excon/excon/issues/529

lanej:
@geemus reflecting upon this, I was attempting to solve the lack of response body data available to the instrumentor for fog-aws requests.  This PR does not solve that issue because [`response_block `](https://github.com/fog/fog-xml/blob/master/lib/fog/xml/sax_parser_connection.rb#L29) execution occurs outside of the Excon middleware stack.

geemus:
Thanks for the update. Let me know if you'd like to discuss further so we can find a different fix for you.

PR #535, opened 2015-10-22
Decompress middleware should handle empty body.

### LINKED ISSUE #533: Decompress middleware fails when response body is empty
When using the Decompress middleware, if the server responds with an empty body the middleware fails with the following error.

    not in gzip format (Zlib::GzipFile::Error)
    excon-0.45.3/lib/excon/middlewares/decompress.rb:24:in `initialize'
    excon-0.45.3/lib/excon/middlewares/decompress.rb:24:in `new'
    excon-0.45.3/lib/excon/middlewares/decompress.rb:24:in `response_call'

I notice several servers respond this way when you combine gzip compression and `If-None-Match` or `If-Modfied-Since`. I would be happy to submit a pull request that checks to make sure the response body isn't empty before attempting decompression.

Thoughts?

aesterline:
This fixes #533.

### DIFF
--- a/lib/excon/middlewares/decompress.rb
+++ b/lib/excon/middlewares/decompress.rb
@@ -12,16 +12,17 @@ module Excon
       end
 
       def response_call(datum)
-        unless datum.has_key?(:response_block)
+        body = datum[:response][:body]
+        unless datum.has_key?(:response_block) || body.nil? || body.empty?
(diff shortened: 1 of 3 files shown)


geemus:
Looks good, sorry for my delay in reviewing. Thanks!

PR #539, opened 2015-12-01
Fix basic auth passwords with special characters does not auth successfully

### LINKED ISSUE #538: Basic auth username/password may be altered before base64 encoded if special characters are used
The basic auth username and password are both `Utils.unescape_form`-ed before being base64 encoded into the request header. The unescaping can get rid of special characters such as `%` and `+` in the password, which renders the authentication to never succeed.

I don't _think_ we should alter the original value of username and password. The Ruby implementation here treats them as-is as well [here](https://github.com/ruby/ruby/blob/ruby_2_2/lib/net/http/header.rb#L428-L430).

stephenchu:
Attempt to fix #538.

Should you accept this PR, I think the same bug might also be lurking in the code path of `#setup_proxy` as well [here](https://github.com/excon/excon/blob/v0.45.4/lib/excon/connection.rb#L463).

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -71,7 +71,7 @@ module Excon
 
       # Use Basic Auth if url contains a login
       if @data[:user] || @data[:password]
-        user, pass = Utils.unescape_form(@data[:user].to_s), Utils.unescape_form(@data[:password].to_s)
+        user, pass = @data[:user].to_s, @data[:password].to_s
         @data[:headers]['Authorization'] ||= 'Basic ' << ['' << user.to_s << ':' << pass.to_s].pack('m').delete(Excon::CR_NL)
       end
 
--- a/tests/basic_tests.rb
+++ b/tests/basic_tests.rb
@@ -96,6 +96,14 @@ Shindo.tests('Excon basics (Basic Auth Pass)') do
       response.status
     end
   end
+
+  with_rackup('basic_auth_complex_password.ru') do
+    tests('with special characters in password').returns(200) do
+      connection = Excon.new('http://127.0.0.1:9292', :user => 'test_user', :password => 'test%FFpassword+plus' )
+      response = connection.request(:method => :get, :path => '/content-length/100')
+      response.status
+    end
+  end
 end
 
 Shindo.tests('Excon basics (Basic Auth Fail)') do
--- a/tests/rackups/basic_auth_complex_password.ru
+++ b/tests/rackups/basic_auth_complex_password.ru
@@ -1,0 +1,14 @@
+require File.join(File.dirname(__FILE__), 'basic')
+
+class BasicAuthComplexPassword < Basic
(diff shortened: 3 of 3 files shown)


geemus:
@stephenchu good catch. I suspect I just copied the unescaping from elsewhere without thinking too critically about it. The base64 encode should turn anything unsafe into a safe character anyway though, so it would probably be redundant in the best of times. I think you are likely also correct about `setup_proxy` repeating the mistake. Would you be up for making a similar fix there? Thanks!

geemus:
Also, I imagine you hit this in usage somewhere, so a release would be helpful? I'd like to get the proxy part also fixed while we are at it, but then I should be able to cut one.

stephenchu:
Sure. I can make a similar fix to `setup_proxy` before you cut a release. I will try EOD today.

geemus:
Sounds great, no huge hurry on my end but definitely appreciate if you are able to do it.

stephenchu:
I have some second thoughts while trying to properly fix the `setup_proxy` escaping.

What if the basic auth creds user is actually a valid (complicated) email, e.g. `user+spam@example.com`. Based on some internet searching, it seems that the proper way to handle the `@` and `+` is to percent-escape them. But there are so many ways to do escaping (CGI, URI, ERB, your `Utils.escape_*` methods, etc.) that I am not confident the impl I currently have even is correct.

Yet, in the case of a complicated password, I think we should choose to do nothing regarding escaping.

I am looking for some guidance here to make sure this change is not a breaking change for some users out there.

geemus:
Ug, good point. It kind of makes me wonder if we should just not-escape at all (and leave it up to the user to do that if necessary). It seems bad in some cases, but at least you would get what you expect. What do you think?

stephenchu:
I think the password should definitely be as-is unaltered.

I also think (right now) that the username part should be some sort of escaped, whichever ones I listed that does percent-escaped, only because I can see having an email `@` sign as a username could be fairly common and if we don't percent escape that (assuming that the existing code before my fix percent-escapes it as well) we could break those users in the wild.

geemus:
Hmm. So, do you think we should just rollback to the old way to get that behavior back and prevent it from breaking for others? Or would that also not provide what you needed?

stephenchu:
How I initially discovered the issue was related to special characters in my password that got inadvertently escaped. My issue was resolved on my end (by changing my password), but in the name of OSS I just wanted to attempt to fix this in master as well. So currently the merged master change (do nothing to username/password would fix it, at the expense of the username potential problem for some users).

If you like my suggestion of escaping username and leave password untouched, then I can probably get to it as another commit.

geemus:
I think maybe for the sake of safety (in terms of breaking changes) and simplicity, I'm going to opt for reverting back to the initial behavior (expecting user to escape it themselves if needed). Maybe we should update docs around this to make that clearer though?

Issue #540, opened 2015-12-03
Cookies are not preserved during redirects


kognate:
I encountered a use case where a response set a cookie and supplied
a redirect location. Excon would redirect correctly, but the new
 location required the cookie value set by the initial response.

There is a PR dealing with this #508 already but it has some things that needed addressing.

geemus:
closing (will defer discussion to #541)

PR #541, opened 2015-12-03
Keep cookies through redirects (and Beyond)

kognate:
This code is based on code from @colindean and
has been extracted into its own middleware.

I encountered a use case where a response set a cookie and supplied
a redirect location. Excon would redirect correctly, but the new
location required the cookie value set by the initial response.

#540 was opened for discussions about this and this PR is based on code found in #508


geemus:
Looks reasonable, did you have particular questions or concerns?

kognate:
I don't think I have any concerns.  I talked to Collin about the code he submitted and I am now on the project he was on when he submitted his changes. We're using a branch based on his code, but would like get back on the main-line Excon.

geemus:
Thanks for the clarification, seems safe to bring in.

colindean:
:beers:

PR #548, opened 2016-02-23
empty host header for unix as per RFC2616

### LINKED ISSUE #547: Host header for HTTP requests over UNIX sockets not empty
Hello,

Today I noticed that with the latest Docker, the `docker-api` gem failed to communicate with the Docker daemon. It always returned a "400 Bad Request" response.

After some digging I found that on the HTTP gem it uses (`excon`) does not set the value for the `Host` header in the request is not *empty* for requests over UNIX sockets.

I reproduced the issue with the following set up on my Linux computer:

1. Use the *socat* tool to open a UNIX listen socket: `socat UNIX-LISTEN:foo.sock STDOUT`
2. Open irb or pry in another window, require 'excon' and do a GET request: `Excon.get 'unix:/foo.sock/whatever', socket: 'foo.sock'`  (Note: The example expects that both commands have the same working directory)

This gives the following output in the `socat` window:

    GET /foo.sock/whatever HTTP/1.1
    User-Agent: excon/0.45.4
    Host: foo.sock

Please note the value for the `Host` header is `foo.sock`. But according to the specification of HTTP it must be empty. Reference [RFC2616, Section 14.26](https://tools.ietf.org/html/rfc2616#section-14.23):

    A client MUST include a Host header field in all HTTP/1.1 request
    messages . If the requested URI does not include an Internet host
    name for the service being requested, then the Host header field MUST
    be given with an empty value.

Thanks for reading!

geemus:
Fixes #547

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -207,7 +207,7 @@ module Excon
       datum[:headers] = @data[:headers].merge(datum[:headers] || {})
 
       if datum[:scheme] == UNIX
-        datum[:headers]['Host']   ||= '' << datum[:socket]
+        datum[:headers]['Host']   = ''
       else
         datum[:headers]['Host']   ||= '' << datum[:host] << port_string(datum)
       end


tlunter:
@geemus :) Wow, quick turn around. Thanks for grabbing that!

Papierkorb:
Yep, that fixes the issue, thanks! Tried my luck at writing a unit test covering this: https://github.com/Papierkorb/excon/commit/fda9d6a712696728bd76290593d0dd16d1be41bc

geemus:
Cool. I'll bring this in. @Papierkorb could you make that into a pull request to ease merging that for me? Thanks!

Papierkorb:
@geemus PR'd https://github.com/excon/excon/pull/551

geemus:
Thanks!

tlunter:
Hi @geemus, could we get a release for this? Thanks!

geemus:
Sure, got distracted and forgot. Will do so shortly.

geemus:
Should be out as v0.46.0

PR #551, opened 2016-02-23
Add unit test for empty Host header in UNIX sockets (PR #548)

Papierkorb:
See https://github.com/excon/excon/pull/548


geemus:
Thanks!

PR #552, opened 2016-02-23
Always call instrumentor with a block

adelcambre:
This changes the instrumentor implementation to pass a block with the
next middleware call in the case of `response_call` and `error_call`.

These calls in the instrumentor now match the `request_call` one.

This is needed to allow calculating the total time for the request.
Before this change the response_call instrumentor was called before the
middleware chain was called, meaning you could not include the time for
receiving and parsing the response in the total time for the call.

NB: If there are any instrumentors out that there that do *not* call the
block if passed in the case of response_call and error_call, the chain
*will* be halted and the response or error will not be handled normally.
I'm not sure if there is a good way around that (there is no way to
determine if the method takes a block (all methods can take a block)),
and some instrumentors *may* want to not call the block in certain
circumstances so ensuring the block always runs doesn't seem correct
either.

### DIFF
--- a/lib/excon/middlewares/instrumentor.rb
+++ b/lib/excon/middlewares/instrumentor.rb
@@ -3,9 +3,12 @@ module Excon
     class Instrumentor < Excon::Middleware::Base
       def error_call(datum)
         if datum.has_key?(:instrumentor)
-          datum[:instrumentor].instrument("#{datum[:instrumentor_name]}.error", :error => datum[:error])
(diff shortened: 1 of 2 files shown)


adelcambre:
On second thought, maybe ensuring the block is always called is the right way to go. This is called an *instrumentor* so it should always behave the same way it would without instrumentation (aside from any instrumentation side effects).

elight:
Paging @geemus

geemus:
Looks good, thanks.

PR #553, opened 2016-02-24
Make it possible for excon to support SSL for non-HTTPS URI schemes.

jessekempf:
Hi,

I'm working on a project that uses Excon middleware to implement a custom protocol that builds atop HTTP. It does _use_ HTTP under the hood, but there's a more complex sequence of operations that have to happen to access a resource. I ran into a bit of excitement trying to get SSL support working -- it's possible to do so by setting the `scheme` parameter to each Excon operation, though that somewhat defeats the purpose of doing a custom protocol implementation using middleware.

Anyhoo, the included patch makes it easy for middleware to do something like this:
```
module MyProtocolClient
  # Excon middleware for talking the MyProtocol protocol.
  class ExconMiddleware < Excon::Middleware::Base
    Excon::DEFAULTS[:middlewares] << self
    Excon::DEFAULTS[:ssl_uri_schemes] << 'my_scheme'

    #
    # BODY
    # OF
    # MIDDLEWARE
    #
  end
end
```
and have Excon use an SSL socket automatically.

Assuming this is OK with you...how should I go about testing this? All the current tests pass as-is, though I'm not sure where I ought to put any feature-specific tests in the current test suite.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -383,7 +383,7 @@ module Excon
       unix_proxy = @data[:proxy] ? @data[:proxy][:scheme] == UNIX : false
       sockets[@socket_key] ||= if @data[:scheme] == UNIX || unix_proxy
         Excon::UnixSocket.new(@data)
-      elsif @data[:scheme] == HTTPS
+      elsif @data[:ssl_uri_schemes].include?(@data[:scheme])
         Excon::SSLSocket.new(@data)
       else
         Excon::Socket.new(@data)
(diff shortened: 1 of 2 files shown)


geemus:
Hey, I'm not sure how best to test this. The existing tests should confirm that it will not impact HTTPS support, and I dunno if it is worth the trouble to get into having other protocol tests (though without them I suppose regressions are more of a threat). Anyway, happy to bring this in and we can see how it looks and re-evaluate on the test front as needed. Thanks!

PR #554, opened 2016-02-26
Make stubs thread-local.

chanks:
Simple tweak to make stubs local to the current thread. This allows for useful stubbing when running tests in parallel (as supported by Minitest).

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -210,9 +210,9 @@ module Excon
       nil
     end
 
-    # get a list of defined stubs
+    # get a list of defined stubs for the current thread
     def stubs
-      @stubs ||= []
+      Thread.current[:_excon_stubs] ||= []
     end
 
     # remove first/oldest stub matching request_params
--- a/tests/middlewares/mock_tests.rb
+++ b/tests/middlewares/mock_tests.rb
@@ -252,5 +252,26 @@ Shindo.tests('Excon stubs') do
     Excon.stubs.clear
   end
 
+  tests("thread-local stubs") do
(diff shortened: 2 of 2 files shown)


geemus:
Thanks!

cobbr2:
Is there any way to override this behavior? We're trying to bring Excon up to date in our builds, but we have many tests that stub from rspec but run from a backend service (usually `ruote` or `capybara`).

geemus:
@cobbr2 I made a quick pull, could you see if that helps? https://github.com/excon/excon/pull/563 Thanks!

cobbr2:
As I noted on #563 : Works great. Thanks! :+1:

geemus:
@chanks heads up, I think I'm actually going to change the default back to global and make thread-local an option. That way both should be supported (and we've already had a couple issues where people were running not-in-parallel and expected the old global behavior). Similar to #563, but with the default being global and thread-local being opt-in. Does that make sense? What do you think?

chanks:
@geemus Thanks for the heads up! The global use-case seems totally reasonable, and setting stubs thread-local is just a single line of configuration for me, which is of course no big deal :smile:

And thanks for merging the PR in the first place! Parallelizable tests are really reassuring for me in speccing a thread-safe API I'm working on, and I'm glad to be able to make them easier for others to work with too.

geemus:
I've made the described change and released in v0.49.0. @chanks you will want to add `Excon.defaults[:stubs] = local` to your test setup when you update, but otherwise I think/hope you should be good to go. Thanks!

Issue #266, opened 2013-09-09
utf-8 encoding issue when uploading a string 


schorsch:
i am reading a file content an put it as string into the reuqest body and get UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)

In connection.rb https://github.com/geemus/excon/blob/master/lib/excon/connection.rb#L176

    if datum[:body].is_a?(String) # write out string body
      socket.write(request << datum[:body]) # write out request + headers + body

you concat the request string (encoded probably utf-8) with whatever encoding comes in from the body, in my case binary and this probably throws the error.

I think making two calls to socket.write could solve the problem. For now i'am saving the string to a tempfile and pass the file handle to excon .. not very elegant

geemus:
I think this was fixed, but do let me know if you are still seeing issues.

schorsch:
Just recently re-checked my hack and removed it, so the issue is resolved for me, thanks for closing

geemus:
Great, thanks for confirming.

Issue #233, opened 2013-05-28
Write request + headers + payload body in single socket.write


mheffner:
In POST requests, for example, the current code performs two writes:
1. socket.write (request + request headers)
2. socket.write(body)

In the case that the request + request headers + body could fit within a single packet, it would be preferable to buffer these requests and put them on the network using a single socket.write call. The current write-write-read pattern can have a negative impact if it triggers a certain aspect of Nagle's algorithm: http://en.wikipedia.org/wiki/Nagle's_algorithm

The various chunking and encoding options that exist for writing the body payload might make this difficult to do in any case but the general body==String case.

PR #557, opened 2016-03-04
Write the first body chunk in the initial packet.

### LINKED ISSUE #556: Excon should combine HTTP headers and body into a single `write()` call
This is the same issue as #233, which was regressed by the fix to #266.

Excon currently suffers from the same issue as outlined in https://gocardless.com/blog/in-search-of-performance-how-we-shaved-200ms-off-every-post-request/; Because the HTTP headers and body are sent in separate packets, the combination of Nagle's algorithm and TCP delayed ACKs can add up to 200ms to request latency.

nelhage:
fixes #556

When #266 was fixed, it caused a regression in the behavior from #233,
which deliberately merged the initial writes to the socket, for network
efficiency and to avoid triggering the pathological behavior caused by
the combination of Nagle's Algorithm and TCP delayed ACKs.

Restore that optimization, in a slightly more general way: Do a
nonblocking read of an initial chunk of data off the provided body, and
merge that with the headers before sending.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -140,9 +140,10 @@ module Excon
 
           # add additional "\r\n" to indicate end of headers
           request << CR_NL
-          socket.write(request) # write out request + headers
 
           if datum.has_key?(:request_block)
+            socket.write(request) # write out request + headers
+
             while true # write out body with chunked encoding
               chunk = datum[:request_block].call
               if FORCE_ENC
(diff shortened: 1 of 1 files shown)


nelhage:
I'm of mixed feelings about the `read_nonblock` approach; I like that it keeps the "we always treat `body` as an IO` simplification introduced in #266, but it also seems like it increases the possible surface area for weird interaction with weird bodies. I'm happy to go back to something that looks more like the pre-#266 code if you'd prefer.

geemus:
Hey, thanks for the details and kicking this off. I started from here and made a couple (hopefully good/reasonable) tweaks here: https://github.com/excon/excon/pull/559. What do you think?

geemus:
adapted into #559

PR #559, opened 2016-03-07
Optimize initial writes

geemus:
possible fix for #556

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -170,10 +170,14 @@ module Excon
               request.force_encoding('BINARY')
             end
             chunk = body.read([datum[:chunk_size] - request.length, 0].max)
-            if FORCE_ENC
-              chunk.force_encoding('BINARY')
+            if chunk
+              if FORCE_ENC
+                chunk.force_encoding('BINARY')
+              end
+              socket.write(request << chunk)
+            else
+              socket.write(request) # write out request + headers
             end
-            socket.write(request << chunk)
 
             while chunk = body.read(datum[:chunk_size])
               socket.write(chunk)
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -140,9 +140,9 @@ module Excon
 
           # add additional "\r\n" to indicate end of headers
           request << CR_NL
-          socket.write(request) # write out request + headers
 
           if datum.has_key?(:request_block)
+            socket.write(request) # write out request + headers
             while true # write out body with chunked encoding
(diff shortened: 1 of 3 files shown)


nelhage:
I did the `read_nonblock` thing because I was worried about a case where we might want to send the request immediately, but don't yet have the body available. I can't actually think of a case in practice where this would be an issue, but if someone were depending on sending the headers immediately (some kind of weird streaming protocol?) this change might cause a regression. I'm happy to declare that use case too weird to support (and users could always use `:request_block` to support sending the headers and then sending the body after a delay).

nelhage:
Otherwise, this should fix the issue so I'm fine with it! I confirmed it works in my test harness.

geemus:
It's not super obvious, but that read actually will be non-block if the settings have non-block turned on (and blocking if it is turned off). So I think it should still support the case you mentioned. You can see where that fits into things here: https://github.com/excon/excon/blob/master/lib/excon/socket.rb#L31 Does that make sense/help?

For the test harness you mention, do you mean it shows that we now avoid the nagel/etc delay? Or just that it works in the general case? I'd be curious if you have ideas on how we might try to protect ourself from re-introducing this in the future.

Thanks!

nelhage:
Ah, awesome on the nonblocking indirection. I think that makes sense.

Yeah, I noticed this via some HTTP benchmarking I was doing, that showed the Nagle/delayed ACK delay; With this change it no longer exhibits the delay.

I haven't looked too deeply at the test suite, but we could imagine injecting a fake `Socket` implementation that records individual `write` calls, and verifying that there's only one for a small request; I'd hope that + a comment in the test explaining why this is important would suffice to prevent another regression.

geemus:
Sounds good. I'll go ahead and bring this in, but if you have a chance to help around regression prevention/testing I'd certainly appreciate the help. Thanks!

lib/excon/connection.rb
```diff
@@ -155,6 +155,8 @@ def request_call(datum)
                 break
               end
             end
+          elsif body.nil?
+            socket.write(request) # write out request + headers
           elsif !body.nil? # write out body
```
Roman2K: Shouldn't that become an `else`? :sweat_smile:

geemus: Shouldn't break anything as elsif, but yeah, else seems easier/clearer. I've fixed it here: https://github.com/excon/excon/commit/24602e2316f0b08734eff4ff72db81662cb26126

Roman2K: Nice

Issue #513, opened 2015-06-19
More specific Errors thrown would be better in some cases


Ladas:
Would be nice if we could have https://github.com/excon/excon/blob/master/lib/excon/errors.rb#L13
defined a subclasses of Excon::Errors::SocketError

so e.g. Excon::Errors::SocketError::BadCertificate, Excon::Errors::SocketError::OpenSSLNotAvailable

Right now we have no other option than look at error message, which is not very reliable solution, rather having specific error classes will be much cleaner.

geemus:
Definitely, I'd certainly welcome any PRs in that direction. Otherwise I'll try to get to it as time allows. Thanks for the suggestion!

sandstrom:
@geemus How about an error named `CertificateError` (since it's hard to know if the certificate is incorrect, or the trusted certs store is lacking)?

Basically what the conditional does already, but with a separate error class.

Also, would you prefer that it inherits from `SocketError`, to avoid breaking backwards compatibility?

Other thoughts?

geemus:
@sandstrom Yeah, `CertificateError` sounds good to me. I think the conditional is probably fine, it would just be a refinement to the error class raised for clarity and potentially better/nicer error handling. Probably best to still inherit from SocketError for backwards-compatibility. Arguably also it is a socket level error from excons perspective, but it does seem slightly weird. Make sense? Thoughts?

geemus:
I think, at least in terms of the certificate error, we are in a better place now.

PR #561, opened 2016-03-17
Make certificate errors a unique class

sandstrom:
@geemus  I've put together a suggestion for how to break certificate errors into its own class. Thoughts, good/bad? :)

A few questions:
- Why do you store the `socket_error` on the error instance? I don't see it being used anywhere.
- And related, why is there a plain error as the default param for the socket error instance? From what I can tell it's always called with an error instance as the first argument (both in tests and in connection.rb).

Please enlighten me, I'm sure there are good reasons to the above, that I haven't grasped.

Related to https://github.com/excon/excon/issues/513

### DIFF
--- a/lib/excon/errors.rb
+++ b/lib/excon/errors.rb
@@ -8,12 +8,29 @@ module Excon
     class SocketError < Error
       attr_reader :socket_error
 
-      def initialize(socket_error=Excon::Errors::Error.new)
-        if socket_error.message =~ /certificate verify failed/
-          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure).")
+      def initialize(socket_error = Excon::Errors::Error.new)
+        if is_a?(CertificateError)
+          super
         else
           super("#{socket_error.message} (#{socket_error.class})")
+          set_backtrace(socket_error.backtrace)
+          @socket_error = socket_error
         end
+      end
+    end
+
+    class CertificateError < SocketError
+      def initialize(socket_error = Excon::Errors::Error.new)
+        msg = "Unable to verify certificate. This may be an issue with the remote host or with Excon." +
+              "Excon has certificates bundled, but these can be customized." +
+              "`Excon.defaults[:ssl_ca_path] = path_to_certs`, " +
+              "`ENV['SSL_CERT_DIR'] = path_to_certs`, " +
+              "`Excon.defaults[:ssl_ca_file] = path_to_file`, " +
+              "`ENV['SSL_CERT_FILE'] = path_to_file`, " +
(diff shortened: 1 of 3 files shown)


geemus:
In both cases, I'm not sure. I think they may just be historical coincidences that slipped in during various refactors (at least looking through blame/history it didn't seem like it was an important and/or explicit choice).

lib/excon/errors.rb
```diff
@@ -8,12 +8,29 @@ class InvalidStub < StandardError; end
...
 
-      def initialize(socket_error=Excon::Errors::Error.new)
-        if socket_error.message =~ /certificate verify failed/
-          super("Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure).")
+      def initialize(socket_error = Excon::Errors::Error.new)
+        if is_a?(CertificateError)
+          super
         else
           super("#{socket_error.message} (#{socket_error.class})")
+          set_backtrace(socket_error.backtrace)
```
geemus: the set_backtrace happens either way, so maybe we can deduplicate and just have it in the socketerror init, but outside this if/else clause?

geemus:
One small point of feedback, otherwise looking good. We might also be able to get rid of the attribute reader, but I'm a little worried it might break something for someone. Getting rid of the default argument is probably safer though, as I imagine it is (at least largely) only used internally. What do you think?

sandstrom:
**Backtrace**

Since Ruby 2.1 there is `Exception#cause` (http://devblog.avdi.org/2013/12/25/exception-causes-in-ruby-2-1/).

My guess is that it was added before 2.1, to be able to retrieve the underlying exception and its backtrace.

Would you be fine with simply dropping the backtrace-overriding?

**Attribute reader**

Searched through a bunch of code (https://searchcode.com/), and could only find one instance where it was used. An older version of `vcr`. I'd say it's safe to drop it too.

https://github.com/vcr/vcr/commit/64a7e0a3269ee365a55cd556892737f1ba1eb7e5#diff-18cfbfcbac77732de990796cb941939dL43

geemus:
For the time being I still expect/intend to support older rubies, so perhaps we should leave the backtrace stuff for the time being. Hmm, maybe we added the reader for VCR or maybe it is just coincidence.

Anyway, how about we limit this current patch to the known good stuff (maybe with the tweak I mention above) and then we can think about further refinements in a separate issue/pull. What do you think?

sandstrom:
@geemus It doesn't break older rubies, only make the backtrace slightly trickier for this specific error instance (in Ruby below 2.1; something which happens in a ton of other libraries in these ruby versions, since far from everyone tries to patch over the backtrace thing). I'd vote for removing it, since it will simplify things quite a lot.

I feel less strongly about dropping the `socket_error` attribute, although I think it's safe (could not find anything except the an version of VCR + its private API). But fine with keeping it!

geemus:
Cool. I'm going to bring this in, as-is, since it represents a simple/clear improvement already. Then if you want to follow up with some of the backtrace stuff that would be solid. I think maybe we can just leave socket_error since it doesn't hurt anything, but maybe we should add an issue suggesting we drop it prior to 1.0 (because we might as well, just don't really have incentive to rush it now and it may actually cause issues for some people, hard to know).

PR #563, opened 2016-03-28
add option to use global stubs

geemus:
You can set the option via `Excon.defaults[:stubs] = :global`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,8 @@ config.before(:all) do
 end
 ```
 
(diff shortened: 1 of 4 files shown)


cobbr2:
Thanks! Works well in my test environment.

geemus:
merged and released in 0.49.0 (note that global is back to being the default, so you shouldn't need to set an option and it should go back to "just working").

Issue #567, opened 2016-05-17
URI Template (RFC 6570) using Addressable


JeanMertz:
I made this small wrapper around `Excon.new` to handle templated URIs:

https://github.com/JeanMertz/excon-addressable

I am working on an `excon-hypermedia` gem to traverse HyperMedia APIs using Excon, I needed this functionality to work with these APIs correctly, that's why I added it.

Just wanted to put this out here, you might be interested in merging this into Excon in your own way, or link to this gem for people interested in this.

_The gem is only tested against the latest Ruby version, breakage will occur on older versions due to using `prepend`._

geemus:
Cool! Would you be up for adding a section in the README to help us start capturing links to related stuff like this? Probably a list of name/link and short description, what do you think?

djberg96:
@JeanMertz Nice, thanks.

@geemus Speaking from experience, I think it should be left as a separate library. Otherwise you increase your maintenance burden. In any case, I think this issue can be closed.

geemus:
@djberg96 sounds reasonable. Thanks!

PR #569, opened 2016-05-20
Add "Plugins and Middlewares" section to README

JeanMertz:
This helps people find interesting tools built on top of Excon. See https://github.com/excon/excon/issues/567#issuecomment-220410188

I also added a TOC, as the README was becoming quite long. I can revert this, if you think it shouldn't be included.


geemus:
Looks great, thanks!

PR #571, opened 2016-05-28
Refactor Excon::Errors into Excon::Error with backwards compatibility

starbelly:
- Move to Excon::Error class
  - Standardize exception class naming convention, examples:
    - Excon::Error.new returns an Excon::Error::Standard
    - Excon::Errors::Error is now Excon::Error::Standard
    - Excon::Errors::InvalidStub is now Excon::Error::InvalidStub
    - Excon::Errors::SocketError is now Excon::Error::Socket
    - Excon::Errors::HTTPStatusError is now Excon::Error::HTTPStatus
  - Create legacy namespace (Excon::Errors) within excon/error.rb and
    alias classes from Excon::Error in Excon::Errors with legacy names for
    backwards compatibility


starbelly:
@geemus I wanted to do more : ) But I figured, best to start with as little change as possible and clean up more over future revisions.

starbelly:
@geemus. Ignore the previous concerns I had. I refactored Excon::Error as module into a class.  I defaulted to a module because I'm obsessed with namespace and collisions for that matter, but as I mentioned not very intuitive. I want to be able to throw and catch Excon::Error and simply can't do that with a module.

geemus:
Looks/sounds good, thanks!

tokengeek:
@starbelly Is there a reason why HTTP status codes stop at 413 in `lib/excon/error.rb`?

I've just had a number of failures surface in fog/fog-brightbox because these have been removed.

geemus:
@tokengeek sorry about that, probably copy/paste error? I'll try to get a fix with that in/out shortly.

tokengeek:
@geemus No worries and thanks.

I wasn't sure when the 500s disappeared, if there was a change somewhere I'd missed.

Bit more confused when I realised 422 had gone.

geemus:
Yeah, the broad fix was to try and have `error` or `errors` and not both for namespaces. Which this accomplishes, but needed to tweak a bunch of stuff (and I think some just accidently got left behind). I think the path there should fix it and I'll cut a version momentarily.

geemus:
Should be fixed in 0.50.1

starbelly:
Oh lawdy! Need to add regression tests for this.

geemus:
Yeah, for now I just did the quick fix. @starbelly also, fwiw, I put the `Error` back on the end of `InternalServerError`. It was weird in general, but that one seems weird without it (due to the way the spec calls it out and stuff).

tokengeek:
Thanks.

After `bundle update` to v0.50.1 and changing the names space `Errors => Error`  this is passing.

The odd case is that everything else passes except `InternalServerError` without the namespace change.

So is there a second needed change related to there and the rename back from `InternalServer`?

starbelly:
@tokengeek Yes, needs to be changed back OR the logic in main needs to be changed. Because it was changed back to InternalServerError InternalServerErrorError is being created in Excon::Errors I hypothesize. @geemus I can update my PR to rename InternalServerError back to InternalServer or I can update the logic to not append "Error" if /Error/.  How would you like to proceed?

geemus:
I think my inclination would be to not add error if it is already there, I prefer not having `error` at all in most of them, but since that is what it is actually called in RFCs and stuff it seems weird to not have it there. Thanks!

starbelly:
@geemus Ok, creating new PR for that.

starbelly:
Right, so I made some bad assumptions about test coverage in excon, my fault totally. Assumptions are a bad thing(tm). Though, this really motivates me to migrate the test suite to rspec and seek 100% coverage (if possible), which we are mulling over the details. I just have to be able to put down the Elixir for a day or two LOL : )

geemus:
No worries, glad Elixir is still going well, am definitely excited about it (but not finding time/focus for it).

PR #572, opened 2016-05-29
optionally allow unstubbed requests to go through

JeanMertz:
I've got a use-case where I want to stub requests, but don't want to block non-stubbed requests from going through.

This is basically the same as VCR's `allow_http_connections_when_no_cassette`, or Webmock's `allow_net_connect!`.

```ruby
Excon.get('http://example.com', :mock => true, :allow_unstubbed_requests => true)
```


geemus:
Thanks!

PR #578, opened 2016-06-28
Don't append "Error" for legacy errors ifclass name ends with Error

starbelly:


geemus:
Thanks, I think/hope this makes sense as the right way (as opposed to dropping error on InternalServerError).

starbelly:
@geemus I think what you said about the spec makes sense IMHO.

Issue #579, opened 2016-07-07
0.50 / 0.50.1 breaks the stickler mirror command


Iristyle:
Stickler, at version `2.4.2` has remained unchanged since 3/13/2015.  However, installing a fresh copy of stickler will pull in `excon ~> 0.37`, which currently pulls `excon` version `0.50.1`.  Attempting to use stickler to mirror a gem will fail with this version of the `excon` gem:

```
stickler mirror --server http://rubygems.delivery.puppetlabs.net ffi --gem-version 1.9.13 --platform=x86-mingw32

Asking http://rubygems.delivery.puppetlabs.net/ to mirror ffi-1.9.13-x86-mingw32 from rubygems.org : /usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/excon-0.50.1/lib/excon/middlewares/expects.rb:10:in `response_call'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/excon-0.50.1/lib/excon/middlewares/response_parser.rb:8:in `response_call'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/excon-0.50.1/lib/excon/connection.rb:389:in `response'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/excon-0.50.1/lib/excon/connection.rb:253:in `request'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/stickler-2.4.2/lib/stickler/repository/remote.rb:296:in `resource_request'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/stickler-2.4.2/lib/stickler/repository/remote.rb:268:in `remote_uri_exist?'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/stickler-2.4.2/lib/stickler/repository/remote.rb:264:in `remote_gem_file_exist?'
/usr/local/opt/rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/stickler-2.4.2/lib/stickler/repository/remote_mirror.rb:15:in `mirror'
[...]

Issue #580, opened 2016-07-07
Excon::Errors::Error isn't always backwards compatible for Excon::Error


penman:
Since `Excon::Errors::Error` is a subclass of `Excon::Error`, code like this won't work as expected:
(This example is from the [VCR](https://github.com/vcr/vcr) test suite.)

```ruby
expect {
  Excon.get("http://localhost:#{VCR::SinatraApp.port}/404_not_200", :expects => 200)
}.to raise_error(Excon::Errors::Error)
```

Would there be any problem with changing `Excon::Errors::Error` to be set straight to `Excon::Error`, rather than being a subclass? It would prevent backwards compatibility issues like this.

geemus:
Thanks for the heads up, we are looking into what I suspect is the same issue here: #579 

So, not entirely sure what the solution will be (it may well be exactly as you suggest), but we should hopefully have a backwards compatibility fix and release soonish.

/cc @starbelly maybe this provides some additional context?

starbelly:
@geemus #581 fixes this

geemus:
Should now be fixed and released in 0.51.0. Thanks!

PR #581, opened 2016-07-07
Alias Errors::Error to Error vs defining new class which inherits

starbelly:
Fixes https://github.com/excon/excon/issues/579 and https://github.com/excon/excon/issues/580


geemus:
@starbelly Thanks for the quick fix!

PR #582, opened 2016-07-08
Fix warnings on open ended deps in gemspec and remove old boiler plate comments

starbelly:
- Lock into to specific versions known to work with ruby 1.9.3-p551
  optimistically which in turn gets rid of warnings during gem build
- Remove support for ruby 1.9.2 and older ( i.e., from travis build )
- remove old boiler plate comments


geemus:
Seems like the tests stall out, for some reason. Any ideas?

starbelly:
@geemus None, but I'm glad I did this :) Keeps hanging on middleware test units. My guess it is container related as I can't reproduce on my end.  Definitely worth looking into further, so will do. 

 Aside from that, what are your thoughts on retiring 1.8.2 and 1.8.3 from builds?

geemus:
We should drop them at some point, almost assuredly when we do 1.0 (which probably should be any-time-now), but maybe we should keep them "just in case" until then? I guess it maybe depends on how troublesome they are and/or become?

starbelly:
Makes sense. They were only troublesome with this PR, that's about all, older rubies not getting along with newer gems, etc. I can close this PR and re-open it later or just leave it open as a TODO. Going to focus on the rspec stuff for the moment, as it seems more important given the pending release of 1.0.

geemus:
Sounds good, thanks.

starbelly:
I backed up the diff and am closing this new. Will re-open soon, trying to get number of issues down.

PR #583, opened 2016-07-11
Intro to rspec

starbelly:
- Addded .rspec and spec directory with default spec helper
 - Added rspec to deps
 - Modified Rakefile for rspec
 - Set default task to execute "tests" then "spec"
 - Alias task test to spec
 - Added initial example which simply checks for an excon version module


geemus:
Thanks!

Issue #585, opened 2016-07-15
Support for Ruby 3.0 frozen strings


kitop:
In Ruby 3.0 string literals will be frozen by default. Right now, in Ruby 2.3 that can be opted in by running with `--enable-frozen-string-literal` or via magic comments.

This blog contains more info: https://wyeworks.com/blog/2015/12/1/immutable-strings-in-ruby-2-dot-3

Is this something `excon` worries about? I was running some test suites with frozen strings enabled and threw an error in `excon`.

I'm more than happy to provide a PR if this is something the maintainers care about. What I have in mind is to add magic comments to _all_ of the files, and fix tests accordingly. If a string is returned to the user, it can cause some trouble if it's frozen and the user modifies it. This would also be something to check.
There might be a very small performance hit, but at the same time most of the strings used are being re-used as they are HTTP headers, so some GC should be reduced because of that.

geemus:
I would be open to getting this fixed. Seems likely to be inevitable anyway (and hopefully the performance isn't too negatively impacted). Happy to discuss.

starbelly:
Not a maintainer, just some guy : ) I dig it. I'm always in favor of getting ready for the future, especially when it doesn't impact the past (legacy).

kitop:
Will start working on this and send a PR as soon as something's ready, to continue discussion.
Thanks!

geemus:
Sounds great, thanks!

On Fri, Jul 15, 2016 at 2:32 PM, Esteban Pastorino <notifications@github.com

geemus:
This was taken care of by #586, just forgot to close this after.

PR #586, opened 2016-07-18
Support for frozen string literals

kitop:
In Ruby 3.0 string literals will be frozen by default. This commit adds
the `# frozen_string_literal: true` magic comment to all `lib` files,
emulating that behavior.

Most `<<` operators where changed by string interpolation or `+`, this
should present little to no performance impact.

Tests and specs were intentionally not changed, as they are being rewritten to use rspec, and seemed like it may be better to change it while moving it. This commit tries to do the minimum to start supporting frozen strings.

From: https://github.com/excon/excon/issues/585

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module Excon
   class Connection
     include Utils
@@ -72,10 +73,10 @@ module Excon
       # Use Basic Auth if url contains a login
       if @data[:user] || @data[:password]
         user, pass = Utils.unescape_form(@data[:user].to_s), Utils.unescape_form(@data[:password].to_s)
-        @data[:headers]['Authorization'] ||= 'Basic ' << ['' << user.to_s << ':' << pass.to_s].pack('m').delete(Excon::CR_NL)
+        @data[:headers]['Authorization'] ||= 'Basic ' + ["#{user}:#{pass}"].pack('m').delete(Excon::CR_NL)
       end
 
-      @socket_key = '' << @data[:scheme]
+      @socket_key = @data[:scheme].dup
       if @data[:scheme] == UNIX
         if @data[:host]
           raise ArgumentError, "The `:host` parameter should not be set for `unix://` connections.\n" +
@@ -83,10 +84,10 @@ module Excon
         elsif !@data[:socket]
           raise ArgumentError, 'You must provide a `:socket` for `unix://` connections'
         else
-          @socket_key << '://' << @data[:socket]
+          @socket_key = @socket_key + '://' + @data[:socket]
         end
       else
-        @socket_key << '://' << @data[:host] << port_string(@data)
+        @socket_key = @socket_key + '://' + @data[:host] + port_string(@data)
       end
       reset
     end
@@ -105,7 +106,7 @@ module Excon
         else
           socket.data = datum
           # start with "METHOD /path"
-          request = datum[:method].to_s.upcase << ' '
+          request = datum[:method].to_s.upcase + ' '
           if datum[:proxy] && datum[:scheme] != HTTPS
             request << datum[:scheme] << '://' << datum[:host] << port_string(datum)
           end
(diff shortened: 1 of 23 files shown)


geemus:
I haven't run the benchmarks (https://github.com/excon/excon/blob/master/benchmarks/concat_vs_interpolate.rb) more recently, but in the past there was a performance difference between concat and interpolate. I would almost certainly have interpolated otherwise (concat is a real pain). I'd be reticent to lose all of that, though I understand some changes may need to occur to deal with frozen strings, I worry this might be changing more than needed. What do you think?

kitop:
Just run them on my machine, and concat is still faster:
```
  [concat, interpolate]

  +-------------+----------+
  | tach        | total    |
  +-------------+----------+
  | concat      | 0.000673 |
  +-------------+----------+
  | interpolate | 0.008862 |
  +-------------+----------+
```

But there's a subtle bug in the benchmark where `key` actually grows in size every iteration, so it becomes "Content-Leght: : 100\r\n" the first iteration, then "Content-Leght: : 100\r\nContent-Leght: : 100\r\n" the second one, and so on. While interpolation is creating new strings every time.
To make it a bit more fair, I think it should be changed to something like:
```ruby
  tach('concat') do
    temp = ''
    temp << key << ': ' << value << "\r\n"
  end
```

And when running that, the results are pretty different:
```
  [concat, interpolate]

  +-------------+----------+
  | tach        | total    |
  +-------------+----------+
  | interpolate | 0.000332 |
  +-------------+----------+
  | concat      | 0.000532 |
  +-------------+----------+
```

Much more closer.

As you said, some of this changes have to happen for immutable strings. For the other ones, I'm not completely sure about the impact, based on that last change to the benchmark. But I may be missing something there, what are your thoughts?

geemus:
Good catch on the benchmark growth, I must have missed that. They do seem much closer there (I was worried when there was more like a 10x difference).

kitop:
Also, just noticed that the `interpolate` part is missing one interpolation:
```ruby
  tach('interpolate') do
    "#{key}: value\r\n"
  end
```

`value` there is not interpolated, and seems like it should. Getting similar results when changing that anyway.

lib/excon/connection.rb
```diff
@@ -72,21 +73,21 @@ def initialize(params = {})
       # Use Basic Auth if url contains a login
       if @data[:user] || @data[:password]
         user, pass = Utils.unescape_form(@data[:user].to_s), Utils.unescape_form(@data[:password].to_s)
-        @data[:headers]['Authorization'] ||= 'Basic ' << ['' << user.to_s << ':' << pass.to_s].pack('m').delete(Excon::CR_NL)
+        @data[:headers]['Authorization'] ||= 'Basic ' + ["#{user}:#{pass}"].pack('m').delete(Excon::CR_NL)
       end
 
-      @socket_key = '' << @data[:scheme]
+      @socket_key = @data[:scheme].dup
```
geemus: I think we can probably do away with this line now to simplify and just have a single interpolating line below. What do you think?

lib/excon/constants.rb
```diff
@@ -29,9 +30,9 @@ module Excon
 
   UNIX = 'unix'
 
-  USER_AGENT = 'excon/' << VERSION
+  USER_AGENT = 'excon/' + VERSION
```
geemus: Should we just interpolate for this and versions?

geemus:
I think I feel more comfortable with this change after your benchmark comments, but still had a couple more nitty gritty comments above. Thanks!

kitop:
Thanks :)

Working on the comments

lib/excon/constants.rb
kitop: Yes, much cleaner I think.

lib/excon/connection.rb
kitop: Makes sense!

kitop:
Comments addressed. Any further feedback is welcome!

geemus:
Looks good, thanks!

PR #589, opened 2016-07-27
Update concat vs interpolation benchmark

kitop:
There was a subtle bug in the benchmark where key actually grows in
size every iteration, so it became "Content-Leght: : 100\r\n" the first
iteration, then "Content-Leght: : 100\r\nContent-Leght: : 100\r\n" the
second one, and so on. While interpolation is creating new strings every
time.
To make it a bit more fair,  temporary variable was created.

Also, in the interpolation part, `value` was not being interpolated.

Benchmark numbers were updated as well.


geemus:
Thanks!

PR #591, opened 2016-08-04
Add Unicorn, Puma, and Exec plugin, support, and tests for Excon::Test::Server

starbelly:
- Added Unicorn plugin for Excon::Test::Server
     - Added Puma plugin fo Excon::Test::Server
     - Added Exec plugin for Excon::Test::Server (i.e., for sinatra, etc.)
     - Added initial tests for Unicorn, Puma, Exec plugins
     - Renamed Rackup|rackup to Webrick|webrick
     - Load Unicorn, Puma, and Exec plugins from Excon::Test::Server by default
     - Added support for URIs in order to better support ipv6
     - Added conditional ops to delete a unix socket if one still exists
       post-shutdown
     - move check_implementation under private to make ruby < 2.0.0 happy
     - Add nested_const_get to stay compat with ruby < 1.9
     - Specify webrick in with_rackup since puma is in the mix now
     - Changed default rspec format to doc instead of progress
     - used shared examples in excon test server spec
     - define get_abs_path to dry up path helpers
     - let rackup_path helper call get_abs_path
     - define unicorn_path and puma_path to support the web server shared example,
       it calls
       rackup_path
     - rename server_path to exec_path and call get_abs_path
     - Skip any tests when unicorn is the server and platform is jruby
     - Add "bundle exec rake spec" to travis.yml
     - Modify task for rspec to take a format argument, defaults to doc
     - use progress format for rspec on travis for less log
     - Added puma to Gemfile
     - Moved unicorn to groups test and devel

Questions: Thought on the Exec namespace? I'm not particularly fond of it, but I was trying to be general, descriptive and concise; so that's what we end up with : )  

Notes: The way I'm doing PRs now is keeping the number of new/deleted/changed files below 10 (perhaps it should be less). If you agree with that practice, perhaps it should be added to the CONTRIBUTING guide.  None the less, only one thing left to do now and that is carry over the actual tests. If I do PRs with less than 10 files added/changed/deleted then I'll need to split that up into 5 PRs. Let me know your thoughts. 

TODO: Clean up,  DRY up, and documentation, will be part of next PR


starbelly:
Yeah, I need to modify Excon::Test::Server to take a host and port as two arguments instead of 'host:port'. Trying to split that up with a regex is just asking for trouble... Will finish this up tomorrow.

geemus:
Looking good.

* Exec namespace seems fine, I can't think of a better alternative off hand and seems good-enough.
* I'm not sure about 10 files rule, but something to keep size of the reviews/diffs down seems good. It's much harder to really pay attention after a certain point otherwise.
* Instead of a regex, maybe just do the `split(':',2)'? I guess it depends if host could have user/pass info.
* Looks like jruby timed out, but I think you were running into that earlier as well?

starbelly:
@geemus Yeah, jruby timed out. I thought it was because of host/port problem but  it seems to be a problem with reading STDOUT/STDERR.
  The problem with splitting goes back to ipv6. I suppose maybe we shouldn't even worry about ipv6 but I figured it would be good idea to go ahead and support it here. If we support ipv6, then we have to accept two forms of an address: 0.0.0.0:0 for ipv4 (host:port) or [ipv6]:port. Where the brackets are required around the address, that's  definitely doable and I wasn't thinking about that last night. We could then allow URIs as such tcp://[ipv6]:9292 or unix://[ipv6]:9292 or tcp://127.0.0.1:9292. And yes we could also allow user@ and so forth. I'm down for that, if you are then I'll get that in this PR. If not, I'll go back to having to specify the host and port in separate arguments. It would be nifty to support unix://user:pass@[ipv6]:port though. We could just use URI to do the parsing of the uri.  Let me know your thoughts.

geemus:
I don't know that much about ipv6 (having not used it much), so the bracket stuff is new to me. Is that common usage? Generally it sounds ok otherwise, just yeah, not very familiar.

starbelly:
@geemus Yeah, described in RFC2732 for urls. With ipv6 there's no way to distinguish the last hextet from the port number and you can't rely on a counting hextets like with dotted quad.

  Ok, so I'll go ahead and get ipv6 support in here and use URI to do the parsing, it works. The host bit is returned with the brackets, but can just strip those off.

geemus:
Sounds good, thanks for clarifying.

starbelly:
Definitely broke my new 10 file round, new rule 15 files ; )

geemus:
Ha! fair enough. Is this ready to merge then (looks reasonable to me)?

starbelly:
Merge away!

geemus:
Thanks!

PR #593, opened 2016-08-20
Allow individual requests to provide a user and password

### LINKED ISSUE #588: Individual requests should allow new/distinct user/pass
Right now the basic auth header is only generated on connection creation (rather than being updated on per-request), as a work around you can explicitly supply the header, but having this just work would be better.

cyberdelia:
Allow individual requests to provide a user and password by recreating the ``Authorization`` if ``user`` and ``password`` are passed.

This will fix #588.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -223,6 +223,11 @@ module Excon
       datum = @data.merge(params)
       datum[:headers] = @data[:headers].merge(datum[:headers] || {})
 
+      if datum[:user] || datum[:password]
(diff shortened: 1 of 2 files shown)


geemus:
Looks good. I think we can remove https://github.com/cyberdelia/excon/blob/dd14d98381c71f0d0d16105c41393435870c79bf/lib/excon/connection.rb#L73 though? As it would now just be duplicated here if/when it is present?

cyberdelia:
@geemus Right this could be removed, will do.

cyberdelia:
Done.

cyberdelia:
@geemus Actually, it will break everything else.

Issue #597, opened 2016-09-13
Allow ssl assets to be strings of already read content


hekaldama:
Excon should allow ssl assets to be strings that have already been read into the program. This is particularly useful when these assets are obtained from secure storage via remote APIs and are already in memory (therefore no longer needing to be `File.read`).

PR to follow.

geemus:
Sounds reasonable, let me know if you have any questions or need help.

hekaldama:
@geemus thanks! I am currently running into `OpenSSL::SSL::SSLContext.new.ca_file` seeming to only support a file and not a already loaded file. Any ideas on that one? Was going to ask a ruby core dev about it when he gets in.

hekaldama:
@geemus seems like it will take some changes to the core in the C code. Do you still want a PR for this even if it is only for the certificate and private key? It will still be helpful if people are using `ssl_verify_peer: false`. LMK

geemus:
Sounds alright. Going to close this for now since 1. we can't fix part of it and 2. the part we can fix we can discuss directly on the relevant PR.

PR #598, opened 2016-09-13
First pass at allowing SSL assets to be already loaded. #597

hekaldama:
This does not allow `:ssl_ca_file` to be already loaded due to how `OpenSSL::SSL::SSLSocket` is implemented. Clients can still benefit from this change if they are using `:ssl_verify_peer => false`.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -65,34 +65,19 @@ module Excon
         ssl_context.verify_mode = OpenSSL::SSL::VERIFY_NONE
       end
 
-      # maintain existing API
-      client_cert_data = if cert_data = @data[:client_cert_data]
-                           cert_data
-                         elsif path = (@data[:client_cert] || @data[:certificate_path])
-                           File.read path
-                         end
-
-      private_key_data = if key_data = @data[:private_key_data]
-                           key_data
-                         elsif path = (@data[:client_key] || @data[:private_key_path])
-                           File.read path
-                         end
-
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
-
-      if client_cert_data && private_key_data
+      if client_cert_data && client_key_data
         ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(client_key_data, client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(client_key_data, client_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
         ssl_context.cert = OpenSSL::X509::Certificate.new(@data[:certificate])
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], client_key_pass)
         end
       end
 
@@ -151,10 +136,37 @@ module Excon
 
     private
 
+    def client_cert_data
+      @client_cert_data ||= if ccd = @data[:client_cert_data]
+                              ccd
+                            elsif path = @data[:client_cert]
+                              File.read path
+                            elsif path = @data[:certificate_path]
+                              warn ":certificate_path is no longer supported and will be deprecated. Please use :client_cert or :client_cert_data"
+                              File.read path
+                            end
+    end
+
     def connect
       # backwards compatability for things lacking nonblock
       @nonblock = HAVE_NONBLOCK && @nonblock
       super
     end
+
+    def client_key_data
+      @client_key_data ||= if ckd = @data[:client_key_data]
+                             ckd
+                           elsif path = @data[:client_key]
+                             File.read path
+                           elsif path = @data[:private_key_path]
+                             warn ":private_key_path is no longer supported and will be deprecated. Please use :client_key or :client_key_data"
+                             File.read path
+                           end
+    end
+
+    def client_key_pass
+      @data[:client_key_pass] || @data[:private_key_pass]
+    end
+
   end
 end
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -66,16 +66,22 @@ module Excon
       end
 
       # maintain existing API
-      certificate_data = @data[:client_cert] || @data[:certificate_path]
-      private_key_data = @data[:client_key] || @data[:private_key_path]
+      client_cert_data = if cert_data = @data[:client_cert_data]
+                           cert_data
+                         elsif path = (@data[:client_cert] || @data[:certificate_path])
+                           File.read path
+                         end
+
+      private_key_data = if key_data = @data[:private_key_data]
+                           key_data
+                         elsif path = (@data[:client_key] || @data[:private_key_path])
+                           File.read path
+                         end
+
       private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
 
-      if certificate_data && private_key_data
-        # Allow clients to pass in already loaded certs and private keys.
-        certificate_data = File.read certificate_data unless certificate_data.gsub(/\n|\r/, '') =~ /-----BEGIN CERTIFICATE.*END CERTIFICATE-----/
-        private_key_data = File.read private_key_data unless private_key_data.gsub(/\n|\r/, '') =~ /-----BEGIN.*PRIVATE KEY.*END.*PRIVATE KEY-----/
-
-        ssl_context.cert = OpenSSL::X509::Certificate.new certificate_data
+      if client_cert_data && private_key_data
+        ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
           ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
         else
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -66,16 +66,20 @@ module Excon
       end
 
       # maintain existing API
-      certificate_path = @data[:client_cert] || @data[:certificate_path]
-      private_key_path = @data[:client_key] || @data[:private_key_path]
+      certificate_data = @data[:client_cert] || @data[:certificate_path]
+      private_key_data = @data[:client_key] || @data[:private_key_path]
       private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
 
-      if certificate_path && private_key_path
-        ssl_context.cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      if certificate_data && private_key_data
+        # Allow clients to pass in already loaded certs and private keys.
+        certificate_data = File.read certificate_data unless certificate_data.gsub(/\n|\r/, '') =~ /-----BEGIN CERTIFICATE.*END CERTIFICATE-----/
+        private_key_data = File.read private_key_data unless private_key_data.gsub(/\n|\r/, '') =~ /-----BEGIN.*PRIVATE KEY.*END.*PRIVATE KEY-----/
+
+        ssl_context.cert = OpenSSL::X509::Certificate.new certificate_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(File.read(private_key_path), private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(File.read(private_key_path), private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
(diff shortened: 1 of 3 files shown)


lib/excon/ssl_socket.rb
```diff
@@ -66,16 +66,20 @@ def initialize(data = {})
       end
 
       # maintain existing API
-      certificate_path = @data[:client_cert] || @data[:certificate_path]
-      private_key_path = @data[:client_key] || @data[:private_key_path]
+      certificate_data = @data[:client_cert] || @data[:certificate_path]
```
geemus: How would you feel about just making a new more explicit key here? ie it might look like:
`certificate_data = @data[:client_cert_data] || @data[:client_cert] && File.read(@data[:client_cert]) || @data[:certificate_path]`.

I think on some level I prefer this (you explicitly opt in to path-or-data, instead of us needing to implicitly figure it out based on contents of the string.

What do you think?

hekaldama: Sounds good to me.

lib/excon/ssl_socket.rb
```diff
@@ -66,16 +66,22 @@ def initialize(data = {})
...
-      certificate_data = @data[:client_cert] || @data[:certificate_path]
-      private_key_data = @data[:client_key] || @data[:private_key_path]
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
+      client_cert_data = if cert_data = @data[:client_cert_data]
+                           cert_data
+                         elsif path = (@data[:client_cert] || @data[:certificate_path])
+                           File.read path
+                         end
+
+      private_key_data = if key_data = @data[:private_key_data]
```
geemus: I think `private_key` is the legacy key here, actually. So this should probably actually be `client_key_data` for consistency. Does that make sense?

geemus: I realize now that the code here makes this unclear due to the local variable names, we should probably change those also (I think they just got left behind). `client_...` is what is documented, so we should standardize on that I think.

hekaldama: Cool. Are you saying we can safely remove `@data[:certificate_path]` and `@data[:private_key_path]`? If so, should we remove `@data[:private_key_pass]` as well?

hekaldama: What I have so far:

```diff
diff --git a/lib/excon/ssl_socket.rb b/lib/excon/ssl_socket.rb
index bbc8ac1..cd20040 100644
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -72,27 +72,27 @@ module Excon
                            File.read path
                          end
 
-      private_key_data = if key_data = @data[:private_key_data]
-                           key_data
-                         elsif path = (@data[:client_key] || @data[:private_key_path])
-                           File.read path
-                         end
+      client_key_data = if key_data = @data[:client_key_data]
+                          key_data
+                        elsif path = (@data[:client_key] || @data[:private_key_path])
+                          File.read path
+                        end
 
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
+      client_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
 
-      if client_cert_data && private_key_data
+      if client_cert_data && client_key_data
         ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(client_key_data, client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(client_key_data, client_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
         ssl_context.cert = OpenSSL::X509::Certificate.new(@data[:certificate])
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], client_key_pass)
         end
       end
```

geemus: Not exactly, I think I was unclear. I think for backwards compatibility it would be good to keep the options, I was just saying the precedence should be a little different. I think it should look something like this (note I'm not writing good code, just trying to show precedence):

```ruby
client_cert_data = @data[:client_cert_data] || File.read(@data[:client_key] || @data[:certificate_path))

client_key_data = @data[:client_key_data] || File.read(@data[:client_key] || @data[:private_key_path])
```

So I think what you have for client_cert_data is pretty much fine as is, it just shuffles around what the name is for the other thing (and the precedence). Does that help clarify?

hekaldama: Thanks for working with me on this and offering feedback :)!

I am unclear how my `if elsif` is different from your `this || (that || that)` snip above wrt precedence. Are you saying that you want it to read easier for people to see what we prefer? E.g. We prefer this (`client_cert_data`), or this (`client_cert`), but this last one (`certificate_path`) we don't really want cause it is the old API.

If so, would something like this be interesting (I am using extract method in order to clean up the body of `initialize`):
```diff
diff --git a/lib/excon/ssl_socket.rb b/lib/excon/ssl_socket.rb
index bbc8ac1..3df7664 100644
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -65,34 +65,23 @@ module Excon
         ssl_context.verify_mode = OpenSSL::SSL::VERIFY_NONE
       end
 
-      # maintain existing API
-      client_cert_data = if cert_data = @data[:client_cert_data]
-                           cert_data
-                         elsif path = (@data[:client_cert] || @data[:certificate_path])
-                           File.read path
-                         end
-
-      private_key_data = if key_data = @data[:private_key_data]
-                           key_data
-                         elsif path = (@data[:client_key] || @data[:private_key_path])
-                           File.read path
-                         end
-
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
-
-      if client_cert_data && private_key_data
+      client_cert_data = cert_data
+      client_key_data = key_data
+      client_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
+
+      if client_cert_data && client_key_data
         ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(client_key_data, client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(client_key_data, client_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
         ssl_context.cert = OpenSSL::X509::Certificate.new(@data[:certificate])
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], client_key_pass)
         end
       end
 
@@ -151,10 +140,33 @@ module Excon
 
     private
 
+    def cert_data
+      if cert_data = @data[:client_cert_data]
+        cert_data
+      elsif path = @data[:client_cert]
+        File.read path
+      elsif path = @data[:certificate_path]
+        warn ":certificate_path is no longer supported and will be deprecated. Please use :client_cert or :client_cert_data"
+        File.read path
+      end
+    end
+
     def connect
       # backwards compatability for things lacking nonblock
       @nonblock = HAVE_NONBLOCK && @nonblock
       super
     end
+
+    def key_data
+      if key_data = @data[:client_key_data]
+        key_data
+      elsif path = @data[:client_key]
+        File.read path
+      elsif path = @data[:private_key_path]
+        warn ":private_key_path is no longer supported and will be deprecated. Please use :client_key or :client_key_data"
+        File.read path
+      end
+    end
+
   end
 end
```
We could also reduce local variables by setting the `*_data` methods to cached instance variables if you would like.
```diff
diff --git a/lib/excon/ssl_socket.rb b/lib/excon/ssl_socket.rb
index bbc8ac1..ea1c301 100644
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -65,34 +65,21 @@ module Excon
         ssl_context.verify_mode = OpenSSL::SSL::VERIFY_NONE
       end
 
-      # maintain existing API
-      client_cert_data = if cert_data = @data[:client_cert_data]
-                           cert_data
-                         elsif path = (@data[:client_cert] || @data[:certificate_path])
-                           File.read path
-                         end
-
-      private_key_data = if key_data = @data[:private_key_data]
-                           key_data
-                         elsif path = (@data[:client_key] || @data[:private_key_path])
-                           File.read path
-                         end
-
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
-
-      if client_cert_data && private_key_data
+      client_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
+
+      if client_cert_data && client_key_data
         ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(client_key_data, client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(client_key_data, client_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
         ssl_context.cert = OpenSSL::X509::Certificate.new(@data[:certificate])
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], client_key_pass)
         end
       end
 
@@ -151,10 +138,33 @@ module Excon
 
     private
 
+    def client_cert_data
+      @client_cert_data ||= if ccd = @data[:client_cert_data]
+                              ccd
+                            elsif path = @data[:client_cert]
+                              File.read path
+                            elsif path = @data[:certificate_path]
+                              warn ":certificate_path is no longer supported and will be deprecated. Please use :client_cert or :client_cert_data"
+                              File.read path
+                            end
+    end
+
     def connect
       # backwards compatability for things lacking nonblock
       @nonblock = HAVE_NONBLOCK && @nonblock
       super
     end
+
+    def client_key_data
+      @client_key_data ||= if ckd = @data[:client_key_data]
+                             ckd
+                           elsif path = @data[:client_key]
+                             File.read path
+                           elsif path = @data[:private_key_path]
+                             warn ":private_key_path is no longer supported and will be deprecated. Please use :client_key or :client_key_data"
+                             File.read path
+                           end
+    end
+
   end
 end
```
I didn't do that though because I am not sure I want people to be able to easily print their `*.pem`s if they are `puts`ing the connection or something. Not sure if that matters or not.

Maybe we could even extract method on the key pass too:
```diff
diff --git a/lib/excon/ssl_socket.rb b/lib/excon/ssl_socket.rb
index bbc8ac1..ea78bd2 100644
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -78,21 +78,19 @@ module Excon
                            File.read path
                          end
 
-      private_key_pass = @data[:client_key_pass] || @data[:private_key_pass]
-
       if client_cert_data && private_key_data
         ssl_context.cert = OpenSSL::X509::Certificate.new client_cert_data
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(private_key_data, client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(private_key_data, client_key_pass)
         end
       elsif @data.key?(:certificate) && @data.key?(:private_key)
         ssl_context.cert = OpenSSL::X509::Certificate.new(@data[:certificate])
         if OpenSSL::PKey.respond_to? :read
-          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey.read(@data[:private_key], client_key_pass)
         else
-          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], private_key_pass)
+          ssl_context.key = OpenSSL::PKey::RSA.new(@data[:private_key], client_key_pass)
         end
       end
 
@@ -151,6 +149,10 @@ module Excon
 
     private
 
+    def client_key_pass
+      @data[:client_key_pass] || @data[:private_key_pass]
+    end
+
     def connect
       # backwards compatability for things lacking nonblock
       @nonblock = HAVE_NONBLOCK && @nonblock
```

geemus: Yes, thank you for being willing to discuss and work through this.

You are right that might logic shouldn't be any different in the big sense, but that the `*_path` variables are the old/deprecated ones. I like the idea of explicitly warning around them. Arguably we could remove them at some point, but we have silently accepted them for so long, that warning for a while seems like what I should have done in the first place if I wanted to actually remove it.

So, I think broadly we are on the same page for this stuff now.

Moving the logic for those to their own methods seems like a good call, it was kind of complicated already and having it inline was messy to say the least.

I don't feel particularly strongly about the instance variable version or local variable version. I guess the instance version might seem slightly better, what do you think?

Thanks!

geemus: Makes sense. Hopefully it will be alright, but we can certainly revisit it if the security portion turns out to be a problem.

geemus:
Thanks!

PR #599, opened 2016-09-19
Rspec time (part 1)

starbelly:
@geemus Despite what this PR shows, I've already carried over all of the basic tests. However, I backed those changes out to keep things simpler. Specifically, I'm interested in focusing on spec/requests/basic_spec and spec/support/shared_examples/shared_examples_for_clients. I wanted to make sure you're in agreement with the use of descriptions, contexts, and examples. The goal as I see it when we run tests it reads a like english. So, when `rspec spec/requests/basic_spec.rb' is run, we get:

```shell
Excon::Connection
  when an explicit uri is passed
    .new
      returns an instance
    when :method is :get and :path is /content-length/100
      #request
        returns an Excon::Response
        Excon::Response
          #status
            returns 200
    when nonblock is true and persistent is false
      when :method is get and :path is /content-length/100
        #request
          returns an Excon::Response
          Excon::Response
            #status
              returns 200
    when nonblock is true and persistent is true
      when :method is get and :path is /content-length/100
        #request
          returns an Excon::Response
          Excon::Response
            #status
              returns 200
    when nonblock is false and persistent is true
      when :method is get and :path is /content-length/100
        #request
          returns an Excon::Response
          Excon::Response
            #status
              returns 200
    when nonblock is false and persistent is false
      when :method is get and :path is /content-length/100
        #request
          returns an Excon::Response
          Excon::Response
            #status
              returns 200

Finished in 5.63 seconds (files took 0.15293 seconds to load)
11 examples, 0 failures
```

 The error_spec, etc needs to follow from this example but as I said I wanted to make sure we were in agreement first. If so, then lots of PRs coming this week. Also, good news, we no longer have to skip any tests when the platform is java, even on unix sockets (thanks puma). 

# Commits

-  Gemfile
   - Require jruby-openssl 0.9
-   Excon::Test::Server:
     - Let timeout default to 20 seconds regardless of platform
     - Use $\ for input record seperator for backwards compat
 -  Excon::Test::Plugin::Server::Puma:
     - Don't perform gsub operation on bind.host in Puma test server plugin
 -  specs:
     - Added basic_spec and first basic examples
     - Added initial spec for Excon::Error in spec/excon/error_spec
     - Added eof_requests_spec in spec/requests (formally bad_tests.rb)
     - Add shared example for streaming clients
     - Moved spec/excon_test_server_spec into spec/excon/test/server_spec
     - added shared_examples_for_test_servers by refactoring out
       the shared_example from server_spec
     - Added test_server_context by refactoring out argument building and
       state variables into a new
     - Added helpers directory with file_path_helpers to boot
     - Modified spec_helper to load helpers and shared examples/context
     - removed boiler plate comments and unused options from spec_helper
     - Skip #headers returns a valid date if jruby and unix socket in
       shared_example_for_clients


lib/excon/test/plugin/server/puma.rb
```diff
@@ -4,7 +4,7 @@ module Plugin
       module Server
         module Puma
           def start(app_str = app, bind_uri = bind)
-            bind.host = bind_uri.host.gsub(/[\[\]]/, '')
+            #bind.host = bind_uri.host.gsub(/[\[\]]/, '')
```
geemus: Is this supposed to still be commented out? Would be good to either uncomment, add an explanation as to why it is commented, or just delete it altogether if it is no longer needed.

geemus:
One small comment, but otherwise seems good. Glad to hear you got jruby calmed down with puma also.

lib/excon/test/plugin/server/puma.rb
starbelly: It was supposed to be deleted. I thought I was justified leaving  it when I hit 325,000 earned galactic credits in battle front, but I guess not.

starbelly:
@geemus Word.

geemus:
Cool, thanks. Bringing this in now.

PR #602, opened 2016-11-10
Fix proxy error handling so it doesn't confuse everyone

calphool:
Wasted an hour today dinking around with the fact that the error handling code for bad proxy environment variables obscures the actual error.  Fixed that.

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -128,7 +128,12 @@ module Excon
       if uri.user
         params[:user] = Utils.unescape_uri(uri.user)
       end
-      Excon::Connection.new(params)
+      begin
+          con = Excon::Connection.new(params)
+      rescue Exception => ex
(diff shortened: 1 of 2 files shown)


geemus:
Sorry to hear you had to fight with that for so long, that is always frustrating.

lib/excon.rb
```diff
@@ -128,7 +128,12 @@ def new(url, params = {})
       if uri.user
         params[:user] = Utils.unescape_uri(uri.user)
       end
-      Excon::Connection.new(params)
+      begin
+          con = Excon::Connection.new(params)
+      rescue Exception => ex
+          print "Unable to establish Excon connection.\nException detail:\n#{ex}\nDid you set your proxy environment variable correctly?\n\n"
```
geemus: If we fix the stuff below, this seems like it might be redundant? What do you think?

geemus:
Added a couple small comments, but otherwise looks good. Thanks!

geemus:
Closing in favor of #624

PR #607, opened 2016-12-01
Correct minor error in README

koenrh:
In order to specify which encodings are accepted in responses, the `Accept-Encoding` request header should be used.


geemus:
Good catch, thanks!

PR #608, opened 2016-12-13
Run tests with ruby 2.3.3

ojiry:
#### Purpose

Run tests with latest stable version.


geemus:
Great idea. Perhaps we should just do `2.3` though? I think that aliases to the newest 2.3, so we would automatically get updates. What do you think?

ojiry:
@geemus I totally agree! Updated to `2.3`.
Thanks 😄

geemus:
Well, it was worth a shot. Seems that travis doesn't like this (and we probably need to be more specific after all). See also: https://github.com/travis-ci/travis-ci/issues/5361. Thanks for your understanding, if you could tweak it back to the specific version I will pull it in once the build passes. Thanks!

ojiry:
Oops! I returned to `2.3.3`.
FYR: Adding `dist: trusty` option travis could use the` 2.3` alias.
cf. https://github.com/otaku-dev/otacop/blob/master/.travis.yml#L2

geemus:
Cool, thanks! I'll keep the trusty option in mind, but this seems fine for the moment anyway.

ojiry:
Thank you for merged!

Issue #604, opened 2016-11-18
Passwords with + (plus character) modified


luxflux:
Any `+` in a password is being removed by excon.

IMHO this happens because of the call to `Utils.unescape_form('a+b')` [here](https://github.com/excon/excon/blob/106f18b6ec7acae7aea773f6621106a1bc753073/lib/excon/connection.rb#L221) because if I change the line to this it works:
```ruby
user, pass = @data[:user].to_s, @data[:password].to_s
```

What's the reason to call `Utils.unescape_form` on the password and username?

geemus:
I don't recall, it may just be historical coincidence (we used an escape function from webrick and ported it in to remove dependency). You can see the function here: https://github.com/excon/excon/blob/master/lib/excon/utils.rb#L82

Actually, upon review, maybe this should use `unescape_uri` instead? (which does the same thing, but without the gsub from `+` to ` `.  The unescape itself is there to ensure that even if it comes in with escapes, we use the raw version before we pass it along. Does that make sense? If that works for you, a patch toward `unescape_uri` for user/pass would be great. Would you be able to help us out there? Thanks!

geemus:
Closed by #614

luxflux:
Oh, I'm sorry.... I missed your reply, @geemus! Thanks for fixing, @raul!

PR #614, opened 2017-01-13
Use Utils.unescape_uri for user/password in Authorization header

raul:
Should fix https://github.com/excon/excon/issues/604 which I hit
while trying to replicate a `curl` authenticated request whose
credentials include `+`. An identical `excon` request was generating
`401`s, and started to work after applying this change.

I haven't found a test associated to the way this header is encoded,
so I've only verified the change with the comparison mentioned above.


geemus:
Thanks!

PR #615, opened 2017-01-25
Add a accept header to request to prevent blocking from waf devices

karmab:
Add a accept header to request to prevent blocking from waf devices ( for instance a mod_proxy with crs_21_protocol_anomalies rules loaded

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -116,7 +116,8 @@ module Excon
     :debug_request        => false,
     :debug_response       => false,
     :headers              => {
-      'User-Agent' => USER_AGENT
+      'User-Agent' => USER_AGENT,
+      'Accept'     =>  '*/*'
     },
     :idempotent           => false,
     :instrumentor_name    => 'excon',


geemus:
Thanks for the patch, setting that as the default value for that header seems reasonable anyway and especially if it can help with the particular issues you are seeing. That said, I believe your patch would cause the accept value to always be `*/*`, even if the user tried to set it to something else. I think a similar approach that should get what you need would be to just add it to the list of default headers (in which case it will show up but can be over-ridden). Does that sound reasonable? It would mean moving the key/value to here: https://github.com/karmab/excon/blob/641594844255f7273b213523dfc09cf617fdb47d/lib/excon/constants.rb#L118

karmab:
makes sense, so i did the change

geemus:
Great. Looks like there is one test that specifically checks the values passed though, which is now breaking because it doesn't expect this value. Could you fix that?

The error can be seen here: https://travis-ci.org/excon/excon/jobs/195575311#L291
And I think the test to be fixed is here: https://github.com/excon/excon/blob/master/tests/middlewares/mock_tests.rb#L99

Thanks!

geemus:
Pulling this in, will fix the test momentarily.

geemus:
@karmab thanks! Sorry for the delay, had totally forgotten about this...

PR #616, opened 2017-01-26
Use secure RubyGems source URI in Gemfile

koenrh:
Noticed that the RubyGems source URI in the Gemfile wasn't using HTTPS yet.


geemus:
Good catch, thanks!

PR #617, opened 2017-01-26
Fix Gemnasium badge, and update a few other URLs to use HTTPS

koenrh:
This fixes the broken Gemnasium badge. While I was at it, I updated a few other URLs to use HTTPS as well (if supported).


geemus:
Cool, thanks!

PR #624, opened 2017-05-19
fix Errors::ProxyParseError to Errors::ProxyParse

hedgesky:
I saw #602 which also fixes this error. But there was no activity on that PR for quite a while; and it rescues `Exception` which is considered to be bad practice.
So open this tiny PR to make people with wrong proxy settings a bit happier 😄


geemus:
Sounds good, thanks!

PR #627, opened 2017-05-31
add client_key_data and client_cert_data to VALID_CONNECTION_KEYS

### LINKED ISSUE #626: client_cert_data is not a valid connection key
Your documentation recommends using client_cert_data:

```
connection = Excon.new('https://example.com',
                       client_cert_data: client_cert_data,
                       client_key_data: client_key_data)
```

But if you do so, you get a warning from https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L373 because neither `client_cert_data` nor `client_key_data` are in `VALID_CONNECTION_KEYS`: https://github.com/excon/excon/blob/master/lib/excon/constants.rb#L64

* It _seems_ that nothing will be affected by this, as the invalid keys aren't dropped. Is that correct?
* Would you like me to PR an update to `VALID_CONNECTION_KEYS` that adds `client_cert_data` and `client_key_data`?

edit:

Those keys do appear to be used here: https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb#L68

llimllib:
closes #626

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -64,8 +64,10 @@ module Excon
   VALID_CONNECTION_KEYS = VALID_REQUEST_KEYS + [
     :ciphers,
     :client_key,
+    :client_key_data,
     :client_key_pass,
     :client_cert,
+    :client_cert_data,
     :certificate,
     :certificate_path,
     :disable_proxy,


geemus:
Awesome, thanks for the fix!

PR #629, opened 2017-06-05
Add optional interval between idempotent retries

acant:
If the :retry_interval parameter is passed then the Idempotent middleware
will sleep for the specified number of seconds between each re-try.

This started as a monkey patch inside a private project.  and I hope that it can be useful to others. 
I think I have updated all the necessary locations:
* added the changed itself to the Idempotent middleware and update the tests
* added :retry_interval to the constant list
* added :retry_interval to the Connection yardoc
* added :retry_interval to the README

Any feedback is greatly appreciated!

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -46,6 +46,7 @@ module Excon
     #     @option params [String] :ciphers Only use the specified SSL/TLS cipher suites; use OpenSSL cipher spec format e.g. 'HIGH:!aNULL:!3DES' or 'AES256-SHA:DES-CBC3-SHA'
     #     @option params [String] :proxy Proxy server; e.g. 'http://myproxy.com:8888'
     #     @option params [Fixnum] :retry_limit Set how many times we'll retry a failed request.  (Default 4)
+    #     @option params [Fixnum] :retry_interval Set how long to wait between retries. (Default 0)
     #     @option params [Class] :instrumentor Responds to #instrument as in ActiveSupport::Notifications
     #     @option params [String] :instrumentor_name Name prefix for #instrument events.  Defaults to 'excon'
     def initialize(params = {})
--- a/lib/excon/middlewares/idempotent.rb
+++ b/lib/excon/middlewares/idempotent.rb
@@ -20,6 +20,9 @@ module Excon
 
         if datum[:idempotent] && [Excon::Errors::Timeout, Excon::Errors::SocketError,
             Excon::Errors::HTTPStatusError].any? {|ex| datum[:error].kind_of?(ex) } && datum[:retries_remaining] > 1
+
+          sleep(datum[:retry_interval]) if datum[:retry_interval]
+
           # reduces remaining retries, reset connection, and restart request_call
           datum[:retries_remaining] -= 1
           connection = datum.delete(:connection)
--- a/tests/middlewares/idempotent_tests.rb
+++ b/tests/middlewares/idempotent_tests.rb
@@ -128,4 +128,37 @@ Shindo.tests('Excon request idempotencey') do
     response.status
   end
 
+  tests("Retry limit and sleep in constructor with socket erroring first 2 times").returns(200) do
+    run_count = 0
+    Excon.stub({:method => :get}) { |params|
+      run_count += 1
+      if run_count <= 2 # First 5 calls fail.
+        raise Excon::Errors::SocketError.new(Exception.new "Mock Error")
+      else
+        {:body => params[:body], :headers => params[:headers], :status => 200}
+      end
+    }
+
+
+    # NOTE: A short :retry_interval will avoid slowing down the tests.
+    response = @connection.request(:method => :get, :idempotent => true, :path => '/some-path', :retry_limit => 3, :retry_interval => 0.1)
+    response.status
+  end
+
+  tests("Retry limit and sleep in constructor with socket erroring first 2 times").raises(Excon::Errors::SocketError) do
+    run_count = 0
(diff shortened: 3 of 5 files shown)


lib/excon/middlewares/idempotent.rb
```diff
@@ -20,6 +20,9 @@ def error_call(datum)
 
         if datum[:idempotent] && [Excon::Errors::Timeout, Excon::Errors::SocketError,
             Excon::Errors::HTTPStatusError].any? {|ex| datum[:error].kind_of?(ex) } && datum[:retries_remaining] > 1
+
+          sleep(datum[:retry_interval]) if datum[:retry_interval] && !datum[:mock]
```
geemus: I wonder if we might want this to occur even in mock mode? If you want to disable it then, you could change the interval, but I think just skipping it might be surprising in some cases. That said, any testing that relates to timing can be rather brittle.

geemus: What do you think?

tests/middlewares/idempotent_tests.rb
```diff
@@ -128,4 +128,34 @@
     response.status
   end
 
+  tests("Retry limit and sleep in constructor with socket erroring first 2 times").returns(200) do
```
geemus: Testing to ensure this doesn't introduce error seems good, but it seems awkward that it doesn't really change the material (timing) change directly. That said, I'm reticent to try and do that as timing based testing can often be so brittle... Thoughts?

lib/excon/middlewares/idempotent.rb
acant: Agreed that timing is tests are a pain. :) And mixing mock behaviour into real code is not a great trade off. Removing the mock and send a 0 internal seems reasonable to me.

Alternatively, what would you think about stubbing #sleep, and checking that it is execute the right number of times?
This would be more like RSpec so it might make sense to convert this test first.

lib/excon/connection.rb
```diff
@@ -46,6 +46,7 @@ def proxy=(new_proxy)
     #     @option params [String] :ciphers Only use the specified SSL/TLS cipher suites; use OpenSSL cipher spec format e.g. 'HIGH:!aNULL:!3DES' or 'AES256-SHA:DES-CBC3-SHA'
     #     @option params [String] :proxy Proxy server; e.g. 'http://myproxy.com:8888'
     #     @option params [Fixnum] :retry_limit Set how many times we'll retry a failed request.  (Default 4)
+    #     @option params [Fixnum] :retry_interval Set how long to wait between retries.
```
acant: @geemus
>  Also, more generally this reminds me of some similar requests I've seen around allowing exponential backoff with retries. That is often seen as a best practice (vs specifically timed retries), so perhaps we could consider just having that as a boolean option instead of integers? Something to consider anyway, what do you think?

I think that it is a good idea, and I did not implement it only because I did not need it immediately. But I'm happy to expand on this but I'm not quite sure about the right way to do it.

One possibility is to add a :retry_multiplier option which defaults to 1. This multiplier would then be used to grow the internal at each iteration.

I want to preserve the existing behaviour, so this command will still not sleep at all:
```
connection.request(:idempotent => true, :retry_limit => 6)
```

This call will still have constant intervals:
```
connection.request(:idempotent => true, :retry_limit => 6, :retry_internal => 10)
```

And this call will back off with the following intervals 1s, 1.5s, 2.25s,  3.4s, 5.1s, 7.6s:
```
connection.request(:idempotent => true, :retry_limit => 6, :retry_internal => 0.5, :retry_multiplier => 1.5)
``` 

Do you think that would match up with previous requests?

geemus: Maybe we could just leave your stuff as-is, and later add another/distinct option, so you could do `connection.request(:idempotent => true, :retry_limit => 6, :exponential_backoff => true)` or something. That might be easiest, rather than trying to more generically do it. What do you think?

lib/excon/middlewares/idempotent.rb
geemus: Yeah, removing it and setting it to 0 in tests seems reasonable for now. stubbing sleep and counting would probably be even better, but I don't think we have to do it now.

lib/excon/connection.rb
acant: I like that idea for keeping things simple. 
For the exponential backoff, does a interval of 1 second and a rate of 1.5 sound good to you? That will end up with the set of of intervals that I described above 1s, 1.5s, 2.25s, 3.4s, 5.1s, 7.6s,...

geemus: Probably, though I would just leave it out for this PR to keep things streamlined.

geemus:
I think there are just 2 things that I'd like to see before pulling this in, hopefully both pretty easy.

1. Make behavior identical in mock/real, but set interval to `0` for tests here.
2. Add to param comment to make it clear that the default interval is `0`.

I think with those changes we can go ahead and bring this in, and we can follow up with exponential related stuff at some later time (and/or just as needed when somebody wants it).

acant:
I will make those updates to the PR, and I am happy to open another for
the exponential backoff.

Also, thank you for the feedback, it has all been very helpful.

geemus:
Sounds great, thanks for taking the time to talk through it all.

acant:
@geemus if you have a moment, I think that this is ready to merge.

geemus:
@acant sorry for the delay (vacation last week), LGTM. Thanks!

geemus:
Now released as part of 0.58.0, thanks again!

PR #631, opened 2017-06-12
Don't override Host when talking to UNIX socket

pbrisbin:
As per the RFC, sending Host="" is the correct behavior when speaking to
a URI that has no host (e.g. a UNIX socket):

> If the requested URI does not include an Internet host name for the
> service being requested, then the Host header field MUST be given with
> an empty value

Unfortunately, some web servers treat that the same as an entirely
omitted Host header, which results in a 400 response. To work around
this, one can fabricate and send any valid Host to which the server will
respond.

For Excon to support this use-case, it needs to respect an explicitly
set Host header, even when talking to a UNIX socket.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -223,7 +223,7 @@ module Excon
       end
 
       if datum[:scheme] == UNIX
-        datum[:headers]['Host']   = ''
+        datum[:headers]['Host']   ||= ''
       else
         datum[:headers]['Host']   ||= datum[:host] + port_string(datum)
       end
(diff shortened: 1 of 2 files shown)


geemus:
@pbrisbin Thanks! Do you need me to cut a release as well?

pbrisbin:
Well, *need* is a strong word, but it'd be appreciated! We're currently monkey-patching `Excon::Connection#request` in our project, and could go on like that for a bit.

geemus:
Released in v0.57.0

PR #634, opened 2017-06-30
FIX: set remote ip to the actual ip we connected

SamSaffron:
This causes me a fair bit of confusion cause the remote IP was not right. While debugging.


geemus:
Good catch. Thanks!

geemus:
Now released in 0.57.1

SamSaffron:
Thanks heaps!

PR #636, opened 2017-07-06
Remove unused error class

shalecraig:
Error class was introduced here: https://github.com/excon/excon/pull/165/files
Usage was removed here: https://github.com/excon/excon/commit/9989457cab47f5d79ddb889d8502ba614e9aeef6

I stumbled upon this while looking at Errors used by the project.


geemus:
Good catch, thanks!

PR #639, opened 2017-08-10
This should be a backtick and not single quote

rohanmendon:
Very minor bug in the error messaging, where it should be backtick and not single quote


geemus:
Oops, good catch. Thanks!

PR #643, opened 2017-08-31
Make thread-safe socket cache per-connection [Fixes #640]

### LINKED ISSUE #640: Excon's socket cache doesn't play well with chunked response callbacks
I'm using Excon via `docker-api`, and if you ask for a stream of events using that library, it provides the event stream in a chunked response, and for each event, it calls a user-provided block.  In that block, you may want to query `dockerd` for details about the object on which the event was triggered.

Unfortunately, that additional request will completely stomp on the event stream response (that's still "in progress"), because Excon only ever uses a single fd for every connection to a specific destination (host/port, or -- in my case -- Unix socket).  Even with separate `Excon::Connection` objects (which `Docker::Connection` creates for every request it makes, even on the same `Docker::Connection`), the problem persists because the default Excon behaviour is to use a per-thread socket cache, which shares sockets between objects in the same thread.

There *is* a workaround I've found that works for my specific case: turn off `thread_safe_sockets`.  In that case, the socket cache is per-connection-object, rather than per-thread, and it works OK *for my specific case*.  It isn't a general solution, however, because if you pass around a single Excon connection to multiple threads that you also use for chunked response processing, you *need* that per-thread cache separation.  (I'm mildly amused that fixing one concurrency problem (multi-threaded access) breaks another sort of concurrency (callbacks), but then, I'm easily amused.)

I suspect that the solution is to make the thread-safe cache *also* per-connection; unfortunately, just tacking `[self.object_id]` onto `Thread.current[:_excon_sockets]` will end up bloating up processes with lots of dead, cached sockets.  I *think* indexing `@_excon_sockets` by the thread ID (ie `@_excon_sockets[Thread.current.object_id][@socket_key]`) will work, but I'm not confident enough about that to submit a PR blind.  If you think it's the right approach, I can put up the trivial PR for it.

mpalmer:
See detailed discussion in #640 for rationale and design discussion.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -408,10 +408,16 @@ module Excon
     end
 
     def sockets
+      @_excon_sockets ||= {}
+
       if @data[:thread_safe_sockets]
-        Thread.current[:_excon_sockets] ||= {}
+        # In a multi-threaded world, if the same connection is used by multiple
+        # threads at the same time to connect to the same destination, they may
+        # stomp on each other's sockets.  This ensures every thread gets their
+        # own socket cache, within the context of a single connection.
+        @_excon_sockets[Thread.current.object_id] ||= {}
       else
-        @_excon_sockets ||= {}
+        @_excon_sockets
       end
     end
 


geemus:
Thanks! Are you able to verify this fixes the docker issue? I presume it should, but would be nice to know for sure. If we feel good on that I can cut a release (though I'm going to be AFK this weekend, I should be able to do it mid next week).

mpalmer:
Yes, I've verified that `docker-api` behaves appropriately both with and without `thread_safe_sockets: false` when I run it with a copy of excon 0.58.0 patched as provided in this PR.

geemus:
Awesome, thanks. As I'll be AFK for the weekend, I don't want to release and run, just in case, but I'll definitely plan to get this out early next week.

geemus:
Released in 0.59.0, thanks again for the discussion and assistance!

Issue #642, opened 2017-08-31
Test suite does not play well with an already-set `http_proxy` env var


mpalmer:
While running the test suite with `rake tests`, I noticed a bunch of spurious failures relating to an IP address not being correct, for example:

    + !!(response.headers['Server'] =~ /^WEBrick/)
            response.headers['Custom'] + returns "Foo: bar"
            response.remote_ip - returns "127.0.0.1"
              expected => "127.0.0.1"
              returned => "192.168.121.1"
            Action? [c,q,r,?]?

I recognise that IP as the one which is set in my `http_proxy` env var.  Sure enough, unsetting the proxy (`http_proxy= rake tests`) works, whilst setting a different (non-existent) address results in failure:

    $ http_proxy=http://192.0.2.42:3128/ rake tests
  
      Excon thread safety
        thread_safe_sockets configuration
          thread_safe_sockets default + returns true
          with thread_safe_sockets set false + returns false
    - long_thread
        connect_write timeout reached (Excon::Error::Timeout)
          /XXX/forks/excon/git/lib/excon/socket.rb:267:in `select_with_timeout'
          /XXX/forks/excon/git/lib/excon/socket.rb:131:in `rescue in block in connect'
          /XXX/forks/excon/git/lib/excon/socket.rb:112:in `block in connect'
          /XXX/forks/excon/git/lib/excon/socket.rb:103:in `each'
          /XXX/forks/excon/git/lib/excon/socket.rb:103:in `connect'
          /XXX/forks/excon/git/lib/excon/socket.rb:29:in `initialize'
          /XXX/forks/excon/git/lib/excon/connection.rb:406:in `new'
          /XXX/forks/excon/git/lib/excon/connection.rb:406:in `socket'
          /XXX/forks/excon/git/lib/excon/connection.rb:101:in `request_call'
          /XXX/forks/excon/git/lib/excon/middlewares/mock.rb:48:in `request_call'
          /XXX/forks/excon/git/lib/excon/middlewares/instrumentor.rb:26:in `request_call'
          /XXX/forks/excon/git/lib/excon/middlewares/base.rb:16:in `request_call'
          /XXX/forks/excon/git/lib/excon/middlewares/base.rb:16:in `request_call'
          /XXX/forks/excon/git/lib/excon/middlewares/base.rb:16:in `request_call'
          /XXX/forks/excon/git/lib/excon/connection.rb:250:in `request'
[...]

PR #644, opened 2017-09-01
Ensure a stray http_proxy env var doesn't ruin everyone's day [Fixes …

mpalmer:
…#642]


geemus:
Ha! Even easier than we suspected, thanks for taking the time to figure that out.

Issue #646, opened 2017-09-13
[RFE] batch requests for pipelining


djberg96:
In some experiments with pipelining large numbers of requests to Microsoft Azure, I would hit timeout errors if I sent too many simultaneous requests. Setting the timeout value may or may not help because sometimes I don't know exactly how many requests I'll be sending, so I'm not sure what I should set the timeout value to, nor what I would do about it anyway if it does timeout.

For example, to get every offer for every publisher for Azure marketplace images, I create a request for each publisher. But, I don't know in advance how many publishers there are going to be. As it turns out, it's now over 500, and that's too much for most connections. The current code looks something like this:

```
publishers.each do |publisher|
  offer_path = File.join(publisher['id'], 'artifacttypes/vmimage/offers')
  path_requests << {:method => :get, :path => offer_path, :query => query}
end

# Ok, now get all offers for all publishers
responses = connection.requests(path_requests)
```
So, what I end up doing is creating batches of 50 or so requests at a time via `each_slice`, and collect all the responses to a single array, and then proceed from there.

While I can certainly do this manually, I thought it would be nice to have a baked-in way to do this, either by allowing a second (integer) argument to the `requests` method, or by creating a `requests_in_batches` method that defaults to, say, 50 requests at a time by default. So, then my code above would just change to:

```
responses = connection.requests_in_batches(path_requests)       # use default of 50
# responses = connection.requests_in_batches(path_requests, 40) # or, set my own value
```

And it would work with a much lower risk of timing out.

geemus:
Thanks for the detailed request. I can certainly see how that might be helpful as a feature. I'll consider it for future enhancement. Thanks!

djberg96:
@geemus Thanks, I should be able to get a PR soon, perhaps this weekend, and you can review it then. It shouldn't be too complicated.

geemus:
Great, thanks! (was traveling last week, so still playing catchup a bit).

PR #647, opened 2017-09-17
Add Connection#requests_in_batches method

djberg96:
This is a simple wrapper for the `Connection#requests` method that allows users to chop up large request sets into smaller batches. What I noticed in practice is that somewhere around 500 simultaneous requests I would get a heavy increase in write time that I couldn't predict, resulting in a timeout error even after setting `:write_timeout` to 5 minutes.

Conversely, I was able to send multiple batches of 250-300 simultaneous requests with the default timeout and receive results in under 2 minutes. So, what I end up doing in practice is slicing up the requests into groups via `each_slice` and collating them into a single result set.

This PR is just a handy wrapper around `Array#each_slice` and `Connection#requests`, with a default slice set at 256.

Addresses https://github.com/excon/excon/issues/646

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,13 @@ connection.requests([{:method => :get}, {:method => :get}])
 By default, each call to `requests` will use a separate persistent socket connection. To make multiple `requests` calls
 using a single persistent connection, set `:persistent => true` when establishing the connection.
 
+For large numbers of simultaneous requests please consider using the `batch_requests` method. This will automatically slice up the requests into batches based on the file descriptor limit of your operating system. The results are the same as the `requests` method, but using this method can help prevent timeout errors.
+
+```ruby
+large_array_of_requests = [{:method => :get, :path => 'some_path'}, { ... }] # Hundreds of items
+connection.batch_requests(large_array_of_requests)
+```
+
 ## Streaming Responses
 
 You can stream responses by passing a block that will receive each chunk.
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -301,6 +301,19 @@ module Excon
       responses
     end
 
+    # Sends the supplied requests to the destination host using pipelining in
+    # batches of @limit [Numeric] requests. This is 256 by default.
+    #   @pipeline_params [Array<Hash>] pipeline_params An array of one or more optional params, override defaults set in Connection.new, see #request for details
+    def requests_in_batches(pipeline_params, limit = 256)
+      responses = []
+
+      pipeline_params.each_slice(limit) do |params|
+        responses.concat(requests(params))
+      end
+
+      responses
+    end
+
     def reset
       if old_socket = sockets.delete(@socket_key)
         old_socket.close rescue nil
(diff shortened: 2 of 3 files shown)


djberg96:
I'm not sure what happened with JRuby, but the CI error appears unrelated.

geemus:
Sorry for the delay, was traveling last week and still trying to play catchup. I think this seems like it is on the right track, but do have some points I'd like to discuss:

1. I wouldn't worry about jruby, it can be a bit flaky for some reason. I restarted it and we'll see if it goes through this time.
2. Naming wise, I'm not sure about `requests_in_batches`, wonder if something like `batch_requests` might be nicer? Or perhaps something else altogether, what do you think?
3. We should probably add at least one test around this. We can probably reuse something very similar to the existing https://github.com/excon/excon/blob/master/tests/pipeline_tests.rb, but maybe with batch size set to 1 (so that there should be 2 batches, and we can compare to the expected non-batched result). Does that make sense?
4. I think it would be helpful to me to see an example of usage, in addition to the comment description. So I suspect that might be helpful to others as well. Maybe we should add a short section to the README describing the new feature with an example. What do you think?

djberg96:
@geemus I looked at pipeline_tests.rb, but cannot tell what it's actually testing, so I'm loathe to just copy and paste.

I also made a couple changes. First, I changed the name to `batch_requests` since you prefer it. Second, I changed the effective default value to a call to `Process.getrlimit`.

starbelly:
@djberg96 it tests persistent connections. 

```ruby
     5       returns(%w{ 1 2 3 4 }, 'connection is persistent') do
  6         connection = Excon.new('http://127.0.0.1:9292', :persistent => true)
  7
  8         ret = []
  9         ret << connection.requests([
 10           {:method => :get, :path => '/echo/request_count'},
 11           {:method => :get, :path => '/echo/request_count'}
 12         ]).map(&:body)
 13         ret << connection.requests([
 14           {:method => :get, :path => '/echo/request_count'},
 15           {:method => :get, :path => '/echo/request_count'}
 16         ]).map(&:body)
 17         ret.flatten
 18       end
```
 First call results in a body of '1', second '2', third 3, fourth 4. If persistence is working then we expect back an array that is equal to [1,2,3,4]... The 'good' server counts the number of request for a given persistent connection. If the connection is not persistent (as in the test below this) we expect [1,2,1,2]. 1 and 2 for the first two requests, and 1 and 2 for the following. The count should never be greater than 2 since the server only increments its count per connection AND request and not just per request (even if it originates from the same source).

djberg96:
@starbelly Thanks.

@geemus Ok, I've pushed up some tests. Please let me know if they're ok.

geemus:
@djberg96 looks good. I think the tests might even be a bit of overkill now, but definitely covers the regression case I was more worried about (we could probably remove all but a couple and still be in good shape in that regard). I think the only other outstanding thing I'd like to see would be an addition to the readme with a description of this and a usage example of how to call and what you might get back. Would you be able to add that also? Thanks!

djberg96:
@geemus Ok, I've updated the README. Please let me know how it looks.

geemus:
Works for me, thanks!

PR #649, opened 2017-10-12
Add Gemfile.lock to .gitignore

djberg96:
I don't think you want to include the Gemfile.lock in your checkins. :)

For me at least a `bundle install` will fail because of eventmachine set at 1.0.4 in the lockfile (current version is 1.2.5).


djberg96:
The failure appears to be JRuby again, btw.

PR #652, opened 2017-10-27
[WIP] Clean up Rakefile

djberg96:
This is a first stab at cleaning up the Rakefile. It eliminates a bunch of stuff in favor of the `bundler/gem_tasks` stuff, which bakes in `build`, `install`, `release` and so on.

I'm not sure if I got the specs and/or Shindo stuff right, please take a look.

I removed the RDoc task, since I don't think anyone really cares any more.

Marked as a WIP for now pending further review and/or suggestions.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,148 +1,18 @@
-require 'rubygems'
-require 'rake'
-require 'date'
-include Rake::DSL
-#############################################################################
-#
-# Helper functions
-#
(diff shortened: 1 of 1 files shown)


geemus:
Looks good to me. I had thought I would try to circle back and do this, so it's a nice surprise that it's already sorted. Thanks!

PR #653, opened 2017-10-30
Remove Gemfile.lock from repo

djberg96:
I should have removed this when I added the Gemfile.lock to the .gitignore file. There's no reason to include the Gemfile.lock file in the repo that I'm aware of. Bundler will regenerate it as needed.


djberg96:
Gemnasium doesn't depend on it, does it? I hope not.

djberg96:
At the moment, no version requirement is set for activesupport. The version that Travis is using (4.04) requires Ruby 1.9.3 or later, and so we get a failure.

I think 1.9.2 should be moved to the allowed failures.

geemus:
@djberg96 sounds like a plan, if you move that and we get back to passing I can bring this in. Thanks!

djberg96:
@geemus Ok, done. Not sure what's up with the current failures.

geemus:
Hmm, those should just be warnings, rather than errors (and are expected). We changed the interface but wanted to still have some tests hit both ways, if I recall, in order to make sure we didn't accidentally break it before we fully removed it. We could probably reduce how much we use it though, as it is definitely very spammy at present.

Looks like the actual failures were timeouts on the servers, which happens sometimes. I think we should be alright here though.

Thanks!

geemus:
Thanks!

PR #654, opened 2017-11-03
Gherkin (demo)

starbelly:
Hey there @geemus so, here's our first gherkin/cucumber example... To run these, `bundle update;cucumber`. I'll do a few more tomorrow perhaps so we can get a better look, but it's pretty straight forward. It's quite clear what's being tested and how it's being tested. Note the re-useable steps... all the steps defined for the one feature can easily be re-used by another or all. Also note you can can opt in for full blown regex or not. 

  You can create the feature file and step definitions all fancy from the get go and have your steps be really  dynamic, but that's not a good way to start IMHO. Besides that, we want to keep the actual steps de-coupled from the feature file(s) as much as possible, naturally. 

P.S. Also note <= 2.1 failing because of certain gems that got updated, can probably work around that and the rest failing because of hang up in middleware tests (not me! 😄 ) and/or travis ci woes.

### DIFF
--- a/excon.gemspec
+++ b/excon.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.test_files       = s.files.select { |path| path =~ /^[spec|tests]\/.*_[spec|tests]\.rb/ }
 
   s.add_development_dependency('rspec', '>= 3.5.0')
+  s.add_development_dependency('cucumber', '>= 3.0.1')
   s.add_development_dependency('activesupport')
   s.add_development_dependency('delorean')
   s.add_development_dependency('eventmachine', '>= 1.0.4')
--- a/features/basic_requests.feature
+++ b/features/basic_requests.feature
@@ -1,0 +1,17 @@
+@requests
+Feature: Basic HTTP requests
+  In order have confidence in Excon and it's more advanced featurees
+  As a user actor
+  I want to ensure the basic functionality works
+
+  Scenario: Basic Request
+    Given a basic web server
+    And a basic excon client
+    When a user gets "/content-length/100"
+    Then the user should receive a response
+    And the status is "200"
+    And the Content-Length header field is "100"
+    And the Content-Type header field is "text/html;charset=utf-8"
+    And the Date header field is a valid date
+    And the Server header field matches "WEBrick"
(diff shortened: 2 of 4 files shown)


starbelly:
Also note <= 2.1 failing because of certain gems that got updated, can probably work around that.

starbelly:
Hmmm, all tests run fine on my laptop... travis keeps timing out on excon middleware test

geemus:
apologies for delays, things have been a bit hectic.

I guess I remain a bit on the fence, it seems an awful lot of indirection for unclear value. I would imagine they could make writing tests more accessible in cases where no new matchers were needed, but I'm not sure how often that actually comes up. Does that make sense?

Maybe you could speak more to the cases where you think this approach would be most helpful so I can better wrap my head around that? Thanks!

starbelly:
I know hectic :) 

I would say being able to write a test without defining a new matcher is a perk, and definitely not the main goal... The main goal is clarity in specification...

```gherkin
 Scenario: Basic Request
    Given a basic web server
    And a basic excon client
    When a user gets "/content-length/100"
    Then the user should receive a response
    And the status is "200"
    And the Content-Length header field is "100"
    And the Content-Type header field is "text/html;charset=utf-8"
    And the Date header field is a valid date
    And the Server header field matches "WEBrick"
    And the remote ip is "127.0.0.1"
```

 It is abundantly clear what the behavior should be. I think a feature spec around something like  https://github.com/excon/excon/pull/647 would prove the case better. Then again, maybe the answer is just to have more comments/docs within the tests themselves.  However, as I've said, if you're not terribly familiar with rspec, I think it's gonna be confusing. Shindo IMHO is less confusing, but out dated (which was part of the reason for the move to rspec).  Of course, it doesn't matter what path you choose, they all have their trade offs. 

 Regardless, I'll get back to finishing up the rspec tests and leave this for the backburner. If at some point we feel differently on the issue, it's quite easy to port rspec tests to cucumber. 

Cheers @geemus

geemus:
@starbelly good points. The readability of specs does often leave something to be desired. Finishing out the specs sounds like a good next step (and we can certainly revisit this at that point). Just getting it all on one test framework would be a nice step forward. Thanks again for your continued help around this!

PR #656, opened 2017-11-29
Add some baked in logging

djberg96:
This adds `Connection#logger` and `Connection#logger=` methods that make it dirt simple to attach logging to Excon. With this PR, users could do something like this:

```
logger = Logger.new("temp.log")
connection = Excon.new("http://www.google.com", :logger => logger)
```

If desired, users can specify a default log using the LoggingInstrumentor directly:

```
Excon::LoggingInstrumentor.logger = logger # Defaults to $stderr if not explicitly set
connection = Excon.new("http://www.google.com", :instrumentor => Excon::LoggingInstrumentor)
```
Since it's a Logger object, most of the formatting responsibility is up to the user and excon doesn't need to worry about anything but basic message output.

On a side note I moved the instrumentors into their own subdirectory just to keep them organized in case you add more instrumentors down the road.


djberg96:
I'm not sure what's going on with Travis.

lib/excon/connection.rb
```diff
@@ -32,6 +32,16 @@ def proxy=(new_proxy)
       @data[:proxy] = new_proxy
     end
 
+    def log
+      if @data[:instrumentor] && @data[:instrumentor].respond_to?(:logger)
+        @data[:instrumentor].logger
+      end
+    end
+    def log=(logger)
```
geemus: Maybe this one should be `logger=`? What do you think?

lib/excon/connection.rb
```diff
@@ -32,6 +32,16 @@ def proxy=(new_proxy)
...
+      Excon::LoggingInstrumentor.logger = logger
+      @data[:instrumentor] = Excon::LoggingInstrumentor
```
geemus: I'm a bit worried that this overwriting instrumentor might be unexpected, though leaving it out makes it much harder to use. Maybe this is a good reason for us to revisit the idea of being able to have an array of instrumentors instead of just one?

lib/excon/instrumentors/logging_instrumentor.rb
```diff
@@ -0,0 +1,53 @@
+require 'logger'
+
+module Excon
+  class LoggingInstrumentor
```
geemus: I wonder a bit if we should iterate on the existing [standard instrumentor](https://github.com/excon/excon/blob/master/lib/excon/standard_instrumentor.rb) to give better behavior around different usage and more flexibility on where things are output, rather than creating a new instrumentor with some overlap? What do you think?

lib/excon/connection.rb
djberg96: I didn't want to confuse people if they wanted to use some other logging mechanism.

lib/excon/connection.rb
djberg96: An array of instrumentors seems reasonable to me, though I don't know what the potential pitfalls are of doing such a thing. Could data get mangled before it gets to my instrumentor, for instance? Just pondering.

lib/excon/instrumentors/logging_instrumentor.rb
djberg96: I thought about that, but wasn't sure if you would object to messing with it. I'd be happy to rework this PR if you prefer that approach.

lib/excon/connection.rb
geemus: Yeah, I guess both `log` and `log=` seemed weird to me if they call logger under the covers. ie Are you getting back a thing to push logs to, or the logs themselves? Does that make sense?

lib/excon/connection.rb
geemus: Good question, I wouldn't expect instrumentors to mutate data, but maybe as part of this we should at least more explicitly document that. What do you think?

lib/excon/instrumentors/logging_instrumentor.rb
geemus: I'm open to this, as long as we can do so in a way that is additive (I wouldn't want to break backwards compatibility). Does that make sense?

lib/excon/connection.rb
djberg96: Ok, I'll update this.

lib/excon/connection.rb
djberg96: Maybe automatically clone any params as part of the interface? If possible, that is. Otherwise, yes, just document that your instrumentors should not mutate data.

lib/excon/instrumentors/logging_instrumentor.rb
djberg96: That might be tricky, but I don't know how many people are relying on $stdout output in production code. Seems like it's purely for debugging, which maybe it should stay, and we should stick with this approach, i.e. separate instrumentors.

lib/excon/connection.rb
geemus: Thanks!

lib/excon/connection.rb
geemus: I considered that, but was concerned about the potential memory bloat...

lib/excon/instrumentors/logging_instrumentor.rb
geemus: Yeah, it's hard to say for sure. Maybe if default logger is still stdout we can make it work, but maybe not. As you suggest it may be more trouble than it's really worth.

lib/excon/connection.rb
djberg96: Ok, updated the method names, and PR description to go with it. Also, I noticed a bug in the way query params were handled, so I updated that as well.

geemus:
Looking good, the last bit from our discussion then, should me make instrumentors work as an array? Backwards compatibility may be a bit of a challenge there, but perhaps worth discussing (though maybe it makes sense to do that work in a separate PR).

djberg96:
If you can make the instrumentors work as an array, please do. But, I think that's a separate PR.

djberg96:
@geemus Do you want me to squash these commits, btw? Or leave them as-is?

geemus:
Those deprecation warnings are expected, though they have been around long enough that maybe we could think about now removing them. No squash needed, thanks for asking though.

geemus:
Thanks!

djberg96:
@geemus Thanks! BTW, when can we expect the next release?

geemus:
Just released 0.60.0, thanks!

PR #665, opened 2018-03-06
Add support for tcp keepalive

ryansch:
This cribs heavily from the way the ruby redis driver handles this issue.  I'm using this with a pool of persistent connections with elasticsearch.

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -151,6 +151,15 @@ module Excon
                            ::Socket::TCP_NODELAY,
                            true)
       end
+
+      if @data[:keepalive]
+        if [:SOL_SOCKET, :SO_KEEPALIVE, :SOL_TCP, :TCP_KEEPIDLE, :TCP_KEEPINTVL, :TCP_KEEPCNT].all?{|c| ::Socket.const_defined? c}
+          @socket.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_KEEPALIVE, true)
+          @socket.setsockopt(::Socket::SOL_TCP, ::Socket::TCP_KEEPIDLE, @data[:keepalive][:time])
+          @socket.setsockopt(::Socket::SOL_TCP, ::Socket::TCP_KEEPINTVL, @data[:keepalive][:intvl])
+          @socket.setsockopt(::Socket::SOL_TCP, ::Socket::TCP_KEEPCNT, @data[:keepalive][:probes])
+        end
+      end
     end
 
     def read_nonblock(max_length)
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -74,6 +74,7 @@ module Excon
     :private_key_path,
     :connect_timeout,
     :family,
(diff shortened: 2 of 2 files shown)


lib/excon/socket.rb
```diff
@@ -151,6 +151,15 @@ def connect
                            ::Socket::TCP_NODELAY,
                            true)
       end
+
+      if @data[:keepalive]
+        if [:SOL_SOCKET, :SO_KEEPALIVE, :SOL_TCP, :TCP_KEEPIDLE, :TCP_KEEPINTVL, :TCP_KEEPCNT].all?{|c| ::Socket.const_defined? c}
+          @socket.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_KEEPALIVE, true)
+          @socket.setsockopt(::Socket::SOL_TCP, ::Socket::TCP_KEEPIDLE, @data[:keepalive][:time])
```
morganick: Do you need any guards around `@data[:kepalive][:time]` where time is `nil`?

ryansch: This is part of a bigger conversation.  The current patch is 'expert mode' that will just blindly pass whatever you give it on to `setsockopt`.  We could instead pull more from the example we have in the redis gem and allow the user to set an integer timeout and do clever things.

geemus: Hey, thanks for starting this discussion. I'm not super familiar with particulars here, so will attempt to engage at broader strokes level, and hopefully the feedback makes sense (but let me know if not). Mostly just some questions though, to help me improve my understanding.

* Are there sockets that would only support a subset of these options, or is it always all or nothing? * Would it make sense for a user to only apply a subset of these settings? If they only work as a full set, are there reasonable defaults we could supply when there is no value?
* Should we warn/error if keepalive has some value, but the socket doesn't support the appropriate options? (I fear the silent failure to use the options may be surprising otherwise).

I would tend to agree that giving the granularity of low level options is probably a good place to start. We can always add a more abstract interface on top to make things easier (but working from more abstract to more concrete is frequently much more awkward).

Hope that helps, certainly happy to continue discussing. Thanks!

ryansch: Honestly I don't know if you'd end up with only partial support.  My understanding is it's an all or nothing affair but redis did it this way and it seems prudent to follow their example.

I think we should emit a warning if keepalive is set but not supported.

geemus: Makes sense to me, partial support seemed unlikely, but it's a lot of checks otherwise. Better safe than sorry perhaps. I agree that emitting a warning if you try to set it and don't have support would be valuable. Would you be up for adding that and then we can see about getting the rest merged? Thanks!

ryansch: I'll grab a time slice this week to work on this.

geemus: Sounds good, just let me know if you need further assistance.

geemus:
Bringing this in, will add warning momentarily.

@ryansch thanks, and sorry for the delay.

Issue #606, opened 2016-11-29
Body encoding is lost during request


chrismwendt:
> Excon wraps the `body` of a request with a `StringIO` instance. This allows Excon to [read the body in chunks](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L166) and send those chunks in the HTTP request. Excon calls `StringIO#binmode` to force binary encoding such that the chunks are not affected by multibyte UTF-8 characters. The problem is that the encoding of the underlying String is also changed by this `binmode` call.

From https://github.com/github/elastomer-client/issues/138

Specifically, the culprit is this `binmode` call: https://github.com/excon/excon/blob/1344c0c4e8fb6ffb26b680867395d693ee57fc07/lib/excon/connection.rb#L156

geemus:
@chrismwendt thanks for the heads up. I suspect a `#dup` would be the probable fix, to ensure we get binmode without modifying the original string? I'm somewhat reticent to do that generally (it's feasible the body string could be quite large, so dup-ing seems potentially dangerous). It may be better to just dup in elastomer, at least if you have a better sense of the size of body you are dealing with to understand the relative safety. Does that make sense? Thoughts?

chrismwendt:
Could excon remember the encoding and set it back to the original after the request? That would avoid duplicating the string.

geemus:
Fair point. I seem to recall some difficulty in getting strings to report their encodings for this purpose across ruby versions, but I could misremember. Would you be interested in working on that to help us out? Thanks!

mltsy:
I just got an error `Encoding::UndefinedConversionError` for a character in a request body (`\xE2`) that I believe is related to this, when trying to use the request body in JSON output.  Could Excon get the character encoding from the content-type header (e.g. from `charset=utf8` if present), and if not present then assume it's `utf8`?  I suppose if this work has already been done, it would make sense to just pull it from the string.  What versions of Ruby are you trying to support?

geemus:
We have historically tried to support everything back through 1.8. I know at some point we will need/want to only do much newer stuff, but that is where we are today (which has a lot to do with why we were not able to find a good working solution as it changed across versions a good deal). We could drop support for older versions, but might want to accompany that with a major release to clearly communicate the breaking change.

mltsy:
[...]

PR #669, opened 2018-03-25
Remove binmode call to keep string encoding

wSnarski:
Removes the `.binmode` call for `String` bodies so that the original encoding of the body is retained after the request is made.

Fix for issue https://github.com/excon/excon/issues/606


geemus:
Thanks!

geemus:
released in v0.62.0

PR #675, opened 2018-05-22
fix header splitting bug

geemus:
Should prevent inadvertent and/or malicious header splitting.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -139,7 +139,10 @@ module Excon
           # add headers to request
           datum[:headers].each do |key, values|
             [values].flatten.each do |value|
-              request << key.to_s << ': ' << value.to_s.gsub(/\r\n/, ' ') << CR_NL
+              if value.to_s.match(/[\r\n]/)
+                raise Excon::Errors::InvalidHeaderValue.new('\r and \n are forbidden')
+              end
+              request << key.to_s << ': ' << value.to_s << CR_NL
             end
           end
 
@@ -185,7 +188,7 @@ module Excon
         end
       rescue => error
         case error
-        when Excon::Errors::StubNotFound, Excon::Errors::Timeout
+        when Excon::Errors::InvalidHeaderValue, Excon::Errors::StubNotFound, Excon::Errors::Timeout
           raise(error)
         else
           raise_socket_error(error)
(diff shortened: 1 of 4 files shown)


geemus:
@ransombriggs I thought headers were split by the combination of `/r/n` not just one or the other? I had originally set it up to do both, but thought a more narrow change was better if it fixed the bug.

I chose replacing with whitespace due to my reading of the RFC. Header continuation describes a header being split over lines using `\r\n`, and that it should be valid for tools to replace with whitespace without changing meaning. So again, this is a narrow/strict interpretation, but I thought that was the safer place to start.

I originally wrote it the way you suggest, albeit substituting in a whitespace instead of blank.

Does that clarify and/or change your opinions?

ransombriggs:
@geemus I now realize I was reading your test incorrectly when I made the comment about guarding against `\r` alone.  I did check the RFC for treatment of just `CR` or `LF` and under [Tolerant Applications](https://tools.ietf.org/html/rfc2616#section-19.3) that they do suggest not being strict.

>   The line terminator for message-header fields is the sequence CRLF.
>   However, we recommend that applications, when parsing such headers,
>   recognize a single LF as a line terminator and ignore the leading CR.

When in the RFC I did find that it does appear valid to [split header fields over multiple lines](https://tools.ietf.org/html/rfc2616#page-31).  It should be noted that this is [obsolete in HTTP/1.1](https://tools.ietf.org/html/rfc7230#section-3.2.4) (with the exception of 

>   Header fields can be
>   extended over multiple lines by preceding each extra line with at
>   least one SP or HT. Applications ought to follow "common form", where
>   one is known or indicated, when generating HTTP constructs, since
>   there might exist some implementations that fail to accept anything

If you really want to support this, I would lean towards erroring out if the header contains `\r\n` without a following `SP` or `HT`.  What is mildly annoying is that they suggest recognizing a single `LF` as the terminator in the tolerant section, so you should apply the same logic for `LF` as `CRLF`.

The [python http library fix](https://hg.python.org/cpython/rev/bf3e1c9b80e9) throws an error on the header value having either `CR` or `LF`, so I think bumping the major version and throwing an error on either would also be valid solution.

geemus:
@ransombriggs changed to error, could you take another look and see if it now matches our discussion? Thanks!

ransombriggs:
@geemus LGTM, ship it

PR #677, opened 2018-06-08
Clarify that username & password should be escaped

ojab:
See https://github.com/excon/excon/pull/539 for the reference.

Took me some debugging time and two account blocks due to too many incorrect attempts to find out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,9 @@ connection = Excon.new('http://geemus.com/', :nonblock => false)
 
 # use basic authentication by supplying credentials in the URL or as parameters
 connection = Excon.new('http://username:password@secure.geemus.com')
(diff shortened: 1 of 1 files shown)


geemus:
Sorry to hear you had troubles with this, it is definitely weird (and had as much to do with trying to avoid breaking changes as anything, if I read/recall correctly). I'm certainly all for updating the docs to clarify for now, and perhaps if/when we make other breaking changes for a major release we should consider changing this also. Thanks!

ojab:
Excon is awesome nevertheless :P

geemus:
@ojab thanks! I'm certainly overall happy with it, despite some small issues around the edges.

Issue #289, opened 2013-10-11
middleware validations


geemus:
Add `valid_connection_keys` and `valid_request_keys` to middleware base class (and override in children). The params validation in connection should then collect these to do validations, so that the list of valid keys can match the configured middlewares instead of matching just default middlewares.

andrew-stripe:
@geemus: are you still interested in this idea? This would be valuable for me since it would allow for writing middleware that accepts custom parameters. 

I think using this to permit new keys is easy but for it would break backwards compatibility to only accept existing keys that are valid for the configured middleware.

geemus:
@andrew-stripe I'd be open to figuring out something here, though we should perhaps discuss specifics a bit more before you get too far in. I'd prefer not to break backwards compatibility, but I'm not entirely opposed to doing so to progress things. Anyway, happy to discuss further.

andrew-stripe:
Along the lines of your original proposal I'd imagine adding `valid_connection_keys` and `valid_request_keys` to the middleware base class. I would then change the logic in `validate_params` to accept the following keys:
  1. Keys returned by the `valid_*_keys` middleware methods for all the provided middleware that respond to these methods.
  2. Keys referenced by Connection itself (stored in a constant somewhere).

I would also add logic to display a warning for keys that are accepted by the current logic but not actually valid for provided middleware explaining the change and how to fix the warnings.

One question is what to do with the existing `VALID_*_KEYS` constants. Do you consider those part of the public interface such that changing or removing them would be a breaking change?

Overall I think this would be a pretty small change.

andrew-stripe:
Mmm, one wrinkle is the fact that the Idempotent middleware references VALID_REQUEST_KEYS:
https://github.com/excon/excon/blob/8fd633e62a5d305ff76ad1f22f10bf015f4ca1c2/lib/excon/middlewares/idempotent.rb#L32

Couple ideas:
* Provide a different way for middleware to trigger retries (e.g. by raising a specific exception or returning a specific value in the datum). This is nice because it would allow custom middleware to implement retries without duplicating the logic in Idempotent (I'm doing this right now).
* Pass `valid_request_keys` down through the datum so it can be read by Idempotent. This is nice because it's very simple.

geemus:
Thanks for the detailed response (and apologies for my delay in responding).

Your described approach sounds reasonable. I wouldn't really consider those constants to be part of the public interface (certainly not purposefully part of it anyway), so changing it should be ok.

[...]

PR #680, opened 2018-06-13
Raise on warnings in tests

metcalf:
Currently Excon tests trigger a `report_warning` in several places that indicate bugs in the code (missing items in the list of valid parameter keys). The tests don't display these warnings so it's difficult to catch this kind of regression. This PR makes it such that tests fail on warnings and updates the code to avoid warnings in tests. This will also facilitate testing the changes in #289 since that will add new warnings and validation behavior.

I used `Excon.set_raise_on_warnings!` to avoid introducing another global variable but I'm open to other implementations.

### DIFF
--- a/spec/helpers/warning_helpers.rb
+++ b/spec/helpers/warning_helpers.rb
@@ -1,0 +1,9 @@
+def silence_warnings
+  orig_verbose = $VERBOSE
+  $VERBOSE = nil
+  Excon.set_raise_on_warnings!(false)
+  yield
+ensure
+  $VERBOSE = orig_verbose
+  Excon.set_raise_on_warnings!(true)
+end
--- a/spec/support/shared_examples/shared_example_for_clients.rb
+++ b/spec/support/shared_examples/shared_example_for_clients.rb
@@ -91,8 +91,10 @@ shared_examples_for 'a basic client' do |url = 'http://127.0.0.1:9292', opts = {
             data = []
             it 'yields with a chunk, remaining length, and total length' do
               expect do
-                conn.request(method: :get, path: '/content-length/100') do |chunk, remaining_length, total_length|
-                  data = [chunk, remaining_length, total_length]
+                silence_warnings do
+                  conn.request(method: :get, path: '/content-length/100') do |chunk, remaining_length, total_length|
+                    data = [chunk, remaining_length, total_length]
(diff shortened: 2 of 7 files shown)


spec/helpers/warning_helpers.rb
```diff
@@ -0,0 +1,9 @@
+def silence_warnings
```
geemus: This seems to just be a duplicate of the same method name in `/tests/test_helper.rb`? If so, we probably only need one or the other?

spec/support/shared_examples/shared_example_for_clients.rb
```diff
@@ -91,8 +91,10 @@
             data = []
             it 'yields with a chunk, remaining length, and total length' do
               expect do
-                conn.request(method: :get, path: '/content-length/100') do |chunk, remaining_length, total_length|
-                  data = [chunk, remaining_length, total_length]
+                silence_warnings do
```
geemus: Could we make this not warning, rather than silencing? (apologies, as I forget exactly what warning might be coming from here).

spec/helpers/warning_helpers.rb
metcalf: Agreed. I wasn't sure what the state of the world was between spec/tests but it looks like spec is the newer thing. I'll keep the helper in spec and have tests require that one.

spec/support/shared_examples/shared_example_for_clients.rb
metcalf: This case is testing `when utilizing deprecated block usage` so we expect to see a warning.

metcalf:
Fixed the code duplication. PTAL @geemus

spec/support/shared_examples/shared_example_for_clients.rb
geemus: Ah, right. Thanks for the reminder.

spec/helpers/warning_helpers.rb
geemus: Yeah, we should probably more explicitly document this. We started a transition toward spec and lost momentum, but the intention was still to land there in the end. At some point we will hopefully get back to that migration and out of this in-between state.

geemus:
Thanks!

PR #685, opened 2018-10-09
Update default retry on errors count in README.md

prog1dev:
Update retry on errors info in README to an actual value of [`Excon::DEFAULT_RETRY_LIMIT`](https://github.com/excon/excon/blob/master/lib/excon/constants.rb#L15) in 
 current master


geemus:
Good catch, thanks!

Issue #689, opened 2019-01-15
Idempotent requests retried on all error types


assembler:
According to this: https://github.com/excon/excon/blob/master/lib/excon/middlewares/idempotent.rb#L41, idempotent requests are retried an all errors. This means that even if server responds with 404 for example, the request will be retried. It would be great if we could specify when to retry requests. 5xx sounds reasonable default.

geemus:
When I was building it, 404 was actually one of the main things I did want to retry. It's a somewhat weird case, but we were working with eventually consistent systems, where there was sometimes a delay before something you created would appear in queries. So, I'm afraid universally changing it would likely be a breaking change.

That said, it might make sense to make it more configurable. You could probably write your own middleware to use instead of this one, for instance, as a way to specifically get what you want. And/or we could discuss what updating the existing one with options might look like.

What do you think?

assembler:
Hey man, thanks for the response! Yes I see that universally changing the default would be a bad idea. Perhaps we can add another option that would control in which instances it is going to perform a retry. Something like this:

```ruby
connection.request(:idempotent => true, :retry_on => [Excon::Error::Server, Excon::Error::Timeout, Excon::Error::Socket])
```

The default would be the same list of errors specified in the idempotent middleware itself. I can make a PR for this if you agree on the concept.

geemus:
Sounds good to me. My only hesitation would be around naming the new key. I was debating between `retry_on` and perhaps `retry_errors`. The second one seems nice in being a bit more specific, but on maybe reads better? I'm kind of on the fence, what do you think? I'm in support of a PR if you'd like to go that route, we might have a little back and forth on details, but broadly this sounds like a nice enhancement. Thanks!

assembler:
Yeah, i share your uncertainty with the name. I prefer `retry_errors` a bit more since it is more expressive so i have used that one in the PR but can easily change it to `retry_on` if you decide to go with that one. Anyway, the PR is ready and specs are passing but i'm unsure if i have structured it properly and not really sure if you test the defaults somewhere more explicitly.

geemus:
Thanks, looks like we are on track, I just added a bit of feedback.

PR #690, opened 2019-01-21
Ability to configure retriable errors

assembler:
Idempotent middleware retries requests on all http errors. This PR enables `retry_errors` option for configuring specific errors on which it should retry (e.g. only 5xx errors). Ref ticket: https://github.com/excon/excon/issues/689#issuecomment-454534684

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -109,9 +109,9 @@ module Excon
     :instrumentor_name => 'Instrumentor',
     :mock => 'Mock',
     :retries_remaining => 'Idempotent', # referenced in Instrumentor, but only relevant with Idempotent
-    :retry_limit => 'Idempotent', # referenced in Instrumentor, but only relevant with Idempotent
+    :retry_errors => 'Idempotent',
     :retry_interval => 'Idempotent',
-    :retry_errors => 'Idempotent'
+    :retry_limit => 'Idempotent' # referenced in Instrumentor, but only relevant with Idempotent
   }
 
   unless ::IO.const_defined?(:WaitReadable)
@@ -152,8 +152,8 @@ module Excon
     :omit_default_port    => false,
     :persistent           => false,
     :read_timeout         => 60,
-    :retry_limit          => DEFAULT_RETRY_LIMIT,
     :retry_errors         => DEFAULT_RETRY_ERRORS,
+    :retry_limit          => DEFAULT_RETRY_LIMIT,
     :ssl_verify_peer      => true,
     :ssl_uri_schemes      => [HTTPS],
     :stubs                => :global,
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -14,6 +14,12 @@ module Excon
 
   DEFAULT_RETRY_LIMIT = 4
 
+  DEFAULT_RETRY_ERRORS = [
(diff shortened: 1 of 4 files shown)


geemus:
Looking good, I'll add a couple minor suggestions presently. Thanks!

lib/excon/constants.rb
```diff
@@ -104,7 +110,8 @@ module Excon
     :mock => 'Mock',
     :retries_remaining => 'Idempotent', # referenced in Instrumentor, but only relevant with Idempotent
     :retry_limit => 'Idempotent', # referenced in Instrumentor, but only relevant with Idempotent
-    :retry_interval => 'Idempotent'
```
geemus: It is very minor/pedantic, but I generally alphabetize keys for things like this. Would you mind moving the retry_errors to the appropriate spot for that? I just find it can make debugging/rereading things smoother in the future (and doesn't usually take much more effort in the present).

tests/middlewares/idempotent_tests.rb
```diff
@@ -161,6 +161,36 @@
     response.status
   end
 
+  tests("Idempotent request with custom error first 3 times").returns(200) do
```
geemus: It would probably be good to also have a test that expects a failure, and raises an error which IS in the default error list, but is NOT in the specified list. I'm pretty confident the test will just pass already, but it may help guard us from regression. Does that make sense?

lib/excon/constants.rb
geemus: I'd ask the same thing for the other instances of the key below, but will just mention it once. Does that make sense? Thanks!

tests/middlewares/idempotent_tests.rb
assembler: Yes, that makes perfect sense since we want to make sure that retry_errors is not complementing the default list but rather completely overriding it. Spec added!

lib/excon/constants.rb
assembler: I have alphabetized retry params where listed. However, i did notice that some of other params are not alphabetized but i really didn't want to touch them in this PR.

geemus: Oops, the others must have slipped through on accident at some point. Thanks for fixing the new one.

geemus:
Thanks!

assembler:
When do you plan to release new version of the gem?

geemus:
Wow, I did not realize it had been so very, very long since I last released. Thanks for the nudge, just pushed out v0.63.0.

assembler:
Yeah, time really flies :) Thanks man!

fwolfst:
Really great, just misses some documentation! 🙏

geemus:
Glad you like it. I would welcome suggestions on how/where to put that documentation to help people find it, or even better I'd love a pull request to improve it if you are up for it. Thanks!

fwolfst:
> Glad you like it. I would welcome suggestions on how/where to put that documentation to help people find it, or even better I'd love a pull request to improve it if you are up for it. Thanks!

@geemus sure thing, see attached PR.

fwolfst:
@assembler Retrying on certain errors (500) is certainly the most typical use case. Do you mind documenting that specific use case (see updated README)?

PR #691, opened 2019-02-05
Ruby 2.6 escapes InvalidURIError output.

voxik:
I observe the following test failure testing against Ruby 2.6:

~~~
  HTTPStatusError request/response debugging
    Excon::Error knows about pertinent errors
    new returns an Error + returns true
    new raises errors for bad URIs + returns true
    new raises errors for bad paths - returns true
      expected => true
      returned => false
    Action? [c,q,r,?]?
~~~

This is apparently caused by https://github.com/ruby/ruby/commit/684cdb4f8340f7a88b00bb91139da74b99ec1147. This PR adds special handling of the output for Ruby 2.6.+

### DIFF
--- a/tests/error_tests.rb
+++ b/tests/error_tests.rb
@@ -35,7 +35,7 @@ Shindo.tests('HTTPStatusError request/response debugging') do
       Excon.new('http://localhost', path: "foo\r\nbar: baz")
       false
     rescue => err
-      err.to_s.include? "foo\r\nbar: baz"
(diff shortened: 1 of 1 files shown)


geemus:
Thanks for digging in to this. I saw the break when I added 2.6, but just hadn't had time to dig into specifics to try and fix it.

voxik:
YAW and sorry for not updating the .travis.yaml. I had no time to update it ;)

geemus:
No worries, that was the easy part!

PR #693, opened 2019-03-04
Add metadata to the gemspec

djberg96:
This PR adds some metadata that gets picked up by rubygems.org for those links on the lower right side (though I don't cover all the ones shown here):

rubygems_links

### DIFF
--- a/excon.gemspec
+++ b/excon.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('json', '>= 1.8.5')
   s.add_development_dependency('puma')
 
-  spec.metadata = {
+  s.metadata = {
     'homepage_uri'      => 'https://github.com/excon/excon',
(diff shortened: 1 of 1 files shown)


excon.gemspec
```diff
@@ -26,4 +26,13 @@ Gem::Specification.new do |s|
   s.add_development_dependency('sinatra-contrib')
   s.add_development_dependency('json', '>= 1.8.5')
   s.add_development_dependency('puma')
+
+  spec.metadata = {
```
geemus: Sounds great. I think this might actually need to be `s.metadata` instead of `spec.metadata` though? (in CI it looks like it is failing on spec being undefined). If you can make that update, I think it should be good to go in terms of merging. Thanks!

djberg96: Oops, fixed.

geemus: No worries, easy mistake, thanks for the quick fix!

geemus:
Thanks for helping us with this extra polish!

PR #700, opened 2019-06-11
fix creating a Proc without a block

ahorek:
fixes
`
    /excon-0.64.0/lib/excon.rb:178: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
`

refs https://github.com/ruby/ruby/commit/9f1fb0a17feb


geemus:
LGTM, thanks!

Issue #354, opened 2014-01-05
stack level too deep when using RedirectFollower


eljojo:
Hi,
when running (with Excon 0.31.0)

``` ruby
middlewares = Excon.defaults[:middlewares] + [Excon::Middleware::RedirectFollower]
url = "http://partners.webmasterplan.com/click.asp?ref=620249&site=7163&type=text&tnb=7&diurl=http://www.buerostuhl24.com/topstar-sitzball-hocker-sitness-half-ball-netz-silber-blau.33360.html"
Excon.get(url, middlewares: middlewares)
```

I get a `SystemStackError: stack level too deep` error message.

The url works fine with `curl -I -L` and with a patched `open-uri` using `open(url).read` gem.

I'm not really sure what might be the cause, but I'm posting it here in case it can help the developers.

Maybe worth noting that I have the `newrelic_rpm` gem installed, which adds the middleware `Excon::Middleware::NewRelicCrossAppTracing`

greysteil:
I'm seeing this with another URL: https://heynemann.github.com/pyvows

geemus:
When I (finally) get around to cutting a v1.0.0, I'll almost assuredly change `:omit_default_port` to be true by default (instead of false by default as it is presently), which will likely avoid this and some related issues. Not sure on my timeline for that exactly, have not had much luck finding time to focus on getting that done.

greysteil:
👍 in this case I have that set, so something else is going on.

Really appreciate everything you do on Excon! If you have an idea for a strategy for a general fix here (presumably passing previous redirects around and checking for circularity) let me know and I can take a look at a fix.

geemus:
Good question. Circularity checking feels like it might be complex, and maybe more trouble than it would be worth (especially given that this doesn't seem to come up super frequently). A simpler approach might be to adopt something similar to the idempotent middleware, where we would count how many times we have passed through the middleware and bail once we reach a configurable limit. It should be less complex, but probably work more or less as-well if configured properly. What do you think?

greysteil:
Sounds perfect - I'll put together a PR.

PR #703, opened 2019-08-04
Limit number of redirects in RedirectFollower

greysteil:
As discussed in https://github.com/excon/excon/issues/354.

I've gone for adding a default of 10 redirects, as I can't imagine anyone wants `SystemStackError` errors raised, but if you'd be happier with keep the currently unlimited default I can change to that.


lib/excon/middlewares/redirect_follower.rb
```diff
@@ -11,9 +23,15 @@ def get_header(datum, header)
       end
 
       def response_call(datum)
+        if datum[:redirects_remaining] <= 0
```
geemus: I feel like we may want to raise an error in this case? The idempotent middleware effectively does this (by allowing the actual error to escape), and I would imagine if you run out of redirects and still have more left, you probably want to know in an exceptional way?

tests/middlewares/redirect_follower_tests.rb
```diff
@@ -30,6 +30,39 @@
   env_restore
 end
 
+Shindo.tests('Excon redirector support with redirect loop') do
+  env_init
+
+  tests("request(:method => :get, :path => '/old').body").returns('redirecting') do
+    Excon.stub(
```
geemus: We should probably have a test that tries to follow more redirects than is allowed to see what happens. (the easiest way, presumably, would be to set the redirect limit to 0 and then get a redirect). Just want to make sure we test both the positive and negative sides.

lib/excon/middlewares/redirect_follower.rb
```diff
@@ -2,6 +2,18 @@
...
     class RedirectFollower < Excon::Middleware::Base
+      def self.valid_parameter_keys
+        [
+          :redirects_remaining,
+          :redirect_limit
+        ]
+      end
+
+      def request_call(datum)
+        datum[:redirects_remaining] ||= datum[:redirect_limit] ||
```
geemus: I like how this keeps this out of the default datum (unlike the idempotent values, which should perhaps be added this way to prevent them from being in datum if idempotent middleware is omitted).

lib/excon/middlewares/redirect_follower.rb
greysteil: Makes sense - updated

tests/middlewares/redirect_follower_tests.rb
greysteil: Added!

greysteil:
Great - updated. I couldn't get `shindo`'s error testing to work, but I noticed you wrote it...!

tests/middlewares/redirect_follower_tests.rb
```diff
@@ -30,6 +30,72 @@
   env_restore
 end
 
+Shindo.tests('Excon redirector support with too many redirects') do
+  env_init
+
+  tests("request(:method => :get, :path => '/old').body").raises(Excon::Errors::TooManyRedirects) do
```
geemus: On further consideration, I think we can omit this test to keep things simple, the redirect loop should be good enough I think. Does that make sense?

tests/middlewares/redirect_follower_tests.rb
```diff
@@ -30,6 +30,72 @@
...
+end
+
+Shindo.tests('Excon redirector support with redirect loop') do
+  env_init
+
+  tests("request(:method => :get, :path => '/old')").raises(Excon::Errors::TooManyRedirects) do
+    Excon.stub(
+      { :path => '/old' },
+      {
+        :headers  => { 'Location' => 'http://127.0.0.1:9292/new' },
```
geemus: For the redirect loop, I think the location should be: `http://127.0.0.1:9292/old` again? That way it redirects to itself and gets stuck?

geemus:
I gave some suggestions on simplifying the tests a little. I'm also having a hard time figuring out why the test isn't catching the raised error and returning success...

geemus:
Ok, I think I found the issue at hand. All errors in connection get rescued, and reraised as socket errors unless they are in a white list: https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L197. I think we would just need to add the redirect error to this list to make sure it passes out directly and it would fix the broken tests. Does that make sense?

greysteil:
Updated. Any idea what I've done to anger `StubNotFound`?

geemus:
I'm not certain, but I do have a guess. Stubs stick around unless you explicitly remove them, so this is probably a race condition where if the tests happen in a certain order it fails due to leftover stubs. The fix should be fairly easy, and it is to add a block that runs after the tests and clears out stubs. Here is an example, which you should be able to pretty much just copy/paste I think: https://github.com/excon/excon/blob/master/tests/middlewares/idempotent_tests.rb#L7. Just let me know if you have questions/concerns, as long as that gets us to green tests, I think we should probably be good to go. Thanks for pointing out the issue and working through the fix with me, I appreciate the help!

greysteil:
My pleasure. I can't get that test passing, though - have tried everything I can think of 😢

geemus:
Hmm, it passes for me locally also, which is not helpful in terms of tracking this down. I'll go ahead and bring in your changes as is and do some extra work to try and hunt down this error. I'm wondering if you are just unlucky in the timing and this was already lurking before your changes.

geemus:
Thanks for your help, if I can get this final test issue ironed out, I'll try to get a release with your enhancement out soon.

greysteil:
Awesome, thanks. A pleasure working with you! :octocat:

geemus:
It appears I was on the right track, but in the wrong place. Looks like the leftover stub was actually in https://github.com/excon/excon/pull/704/commits/c7f2174bfa18fb9952fd9255964247355e14975c. It was passing locally for me because the file order for test runs is different (probably operating system differences). In any event, I think it is cleared up now.

geemus:
Released in v0.66.0. Thanks!

PR #709, opened 2019-11-21
fixed response status check when making a request with a valid proxy …

jasquat:
All calls raise Excon::Errors::ProxyConnectionError when making a call with http_proxy or https_proxy set to a valid address in the environment.

It looks like the issue is that status is a subkey of response and not at the root of the hash:
```
{:expects=>200, :method=>"CONNECT", :response=>{:body=>"", :cookies=>[], :host=>nil, :headers=>{}, :path=>nil, :port=>nil, :status=>200, :status_line=>"HTTP/1.1 200 Connection established\r\n", :reason_phrase=>"Connection established", :remote_ip=>"[IP]", :local_port=>[PORT], :local_address=>"[IP]"}}
```

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -108,7 +108,7 @@ module Excon
 
         # eat the proxy's connection response
         response = Excon::Response.parse(self,  :expects => 200, :method => 'CONNECT')
-        if response[:status] != 200
+        if response[:response][:status] != 200
           raise(Excon::Errors::ProxyConnectionError.new("proxy connection is not exstablished"))
         end
       end


geemus:
Thanks for catching that and the quick fix.

geemus:
Released in 0.69.1.

jasquat:
> Thanks for catching that and the quick fix.

No problem. Thanks for the quick merge.

geemus:
I really wish we had some better testing around this to catch these kinds of issues. But in my attempts in the past I never had much luck emulating this behavior in the test stack, and don't personally use proxies enough to feel confident about manually testing (and manual testing is brittle). If you have any thoughts/advise for how we could catch this better and/or avoid regressions I would definitely welcome the assistance.

PR #715, opened 2020-03-16
embed response from proxy in ProxyConnectionError

### LINKED ISSUE #714: Return proxy response in ProxyConnectionError
It would be helpful to have access to the complete response from the proxy when a non-200 is returned. Currently, `ProxyConnectionError` does not return the response and clients are unable to infer the cause of the proxy failure.

cds2-stripe:
This PR extends `ProxyConnectionError` to include the CONNECT response from the proxy. Doing so allows clients to better understand the cause of the proxy failure. 

This closes #714.

### DIFF
--- a/lib/excon/error.rb
+++ b/lib/excon/error.rb
@@ -52,11 +52,12 @@ or:
     class ResponseParse < Error; end
 
     class ProxyConnectionError < Error
-      attr_reader :proxy_response
+      attr_reader :request, :response
 
-      def initialize(msg, proxy_response = nil)
+      def initialize(msg, request = nil, response = nil)
         super(msg)
-        @proxy_response = proxy_response
+        @request = request
+        @response = response
       end
     end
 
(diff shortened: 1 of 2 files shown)


geemus:
Hey, thanks, this seems like a great addition. I'll add a bit of feedback presently.

lib/excon/error.rb
```diff
@@ -50,7 +50,16 @@ class InvalidHeaderKey < Error; end
     class InvalidHeaderValue < Error; end
     class Timeout < Error; end
     class ResponseParse < Error; end
-    class ProxyConnectionError < Error; end
+
+    class ProxyConnectionError < Error
```
geemus: Could we change this interface to look like the HTTPStatus errors? I don't think it makes sense to derive from that class, but I do think it probably makes sense to have `:response` as the attribute, and probably `:request` as well for consistency. Does that make sense? What do you think?

Here is the HTTPStatus code I'm referring to: https://github.com/excon/excon/blob/c39c6a2775ad3931a9ece2e847eb7d79a37ba048/lib/excon/error.rb#L67

cds2-stripe: 🤦‍♂ that makes perfect sense. We could also use some of the information in the original request. Appreciate the quick response!

geemus:
Thanks for the report and fix. Released in 0.73.0.

PR #720, opened 2020-06-18
Excon::Socket.readline to use buffered reads

kryzthov-stripe:
`Excon::Socket.readline` currently reads 1 byte at a time,
by performing one syscall for each byte (through `::Socket::read_nonblock`).
This change proposes to use the private `Excon::Socket.read_nonblock` primitive
instead, to benefit from the read buffer.

Tests & specs seem to run fine with this change.

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -60,18 +60,8 @@ module Excon
     def readline
       return legacy_readline if RUBY_VERSION.to_f <= 1.8_7
       buffer = String.new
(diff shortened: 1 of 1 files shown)


geemus:
Sorry for the delay, I was just a bit anxious as it feels like a bigger/more risky change. I think it should be fine though, as you suggest. Thanks for noticing this, it is nice to remove some of the non-block error handling duplication as well.

PR #721, opened 2020-07-02
Remove unnessary #dup string in Utils

wittawasw:
I have reviewed some changes in Excon gem before upgrading today and found that these `str.dup` part in `Excon::Utils` is not necessary since we've already check frozen string inside `binary_encode` method

### DIFF
--- a/lib/excon/utils.rb
+++ b/lib/excon/utils.rb
@@ -94,31 +94,27 @@ module Excon
     # Splits a header value +str+ according to HTTP specification.
     def split_header_value(str)
       return [] if str.nil?
-      str = str.dup.strip
-      str = binary_encode(str)
+      str = binary_encode(str.strip)
       str.scan(%r'\G((?:"(?:\\.|[^"])+?"|[^",]+)+)
                     (?:,\s*|\Z)'xn).flatten
     end
 
     # Escapes HTTP reserved and unwise characters in +str+
     def escape_uri(str)
-      str = str.dup
       str = binary_encode(str)
       str.gsub(UNESCAPED) { "%%%02X" % $1[0].ord }
     end
 
     # Unescapes HTTP reserved and unwise characters in +str+
     def unescape_uri(str)
-      str = str.dup
       str = binary_encode(str)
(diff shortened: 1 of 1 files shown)


geemus:
I think the dups may actually serve different purposes in the two places, though my memory is a bit fuzzy at this point.

In `binary_encode` the dup is to get a non-frozen string, as you suggest. In the other parts of `Excon::Utils` however, I think they are there so that the original string will not be mutated (which might be unexpected/surprising). The dupped string that it passes to `binary_encode` should also no longer be frozen, so things passing through both should only be dupped once.

Does that make sense? I think to avoid mutating the strings in those methods we probably still need these.

wittawasw:
I have a similar thought,too. That the `.dup` inside other methods than the one that's inside `binary_encode` is to prevent mutation in original string. 

So, I tried writing a snippet to prove that `.dup` inside those methods are really uncessary by using `.gsub` and various string methods and I saw that it wouldn't changed. But just try it again with `force_encoding` and the mutated string encoding is really there to see. 

I've rarely have a need to use `.force_encoding` myself. So, I've never noticed before that `.force_encoding` will mutate original string itself.

The snippet is below in case anyone having the same misunderstanding as me.

```
def escape_str(str)
  str.force_encoding('BINARY').gsub(/ss/, '')
end

# original string
str = '123ss456ss789'
=> "123ss456ss789"
str.encoding
=> #<Encoding:UTF-8>


# mutated string from method.
str2 = escape_str(str)
str2
=> "123456789"
str2.encoding
=> #<Encoding:ASCII-8BIT>


# unchanged original string at first glance but the encoding is actually changed.
str
=> "123ss456ss789"
str.encoding
=> #<Encoding:ASCII-8BIT>

```

geemus:
Yeah, I suspect in most cases it would be hard to notice and wouldn't cause problems, but there might be outliers. So I'd prefer to play it safe and avoid introducing what might otherwise be a backwards incompatible breaking change on accident. Thanks for offering the suggestion and discussing it though, certainly let me know if you see other possibilities for improvement.

PR #722, opened 2020-07-17
Support ssl_verify_hostname parameter to use OpenSSL::SSL::SSLContext#verify_hostname=

cosmo0920:
Currerntly, `Excon` does not handle `OpenSSL::SSL::SSLContext#verify_hostname=` method on `Excon::SSLSocket` class and does not provide way to handle whether handling [`verify_hostname` option in ruby openssl](https://github.com/ruby/ruby/pull/2858) or not.
Is this planned feature?

Signed-off-by: Hiroshi Hatake <cosmo0920.oucc@gmail.com>

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -95,6 +95,7 @@ module Excon
     :ssl_verify_callback,
     :ssl_verify_peer,
     :ssl_verify_peer_host,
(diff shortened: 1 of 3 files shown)


geemus:
Looks great, thanks!

cosmo0920:
Thank you!

PR #723, opened 2020-09-25
Make Host header be always the first in requests

andrepiske:
I observed that some servers (Cloudflare, for instance) will fail the request or will respond with redirects to the same URL if the Host header comes too late in the request.

I think this actually make sense from a security perspective, as having Host late in the request would mean a server with virtual hosts would have to buffer everything before the Host header before deciding where the request will land and whether the virtual host is even valid.

[RFC 7230 section 5.4](https://tools.ietf.org/html/rfc7230#section-5.4) supports this, stating that:

> Since the Host field-value is critical information for handling a
> request, a user agent SHOULD generate Host as the first header field
> following the request-line.

I propose here this changes that always lifts the Host header to be the first one sent. By looking at the previous lines, the `if` condition shouldn't alter semantics as I don't see how `datum[:headers]['Host']` could be `nil`.

What do you think? Happy to discuss solutions.

Would be also nice if someone could point me into which file I could add tests/specs :)

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -252,6 +252,8 @@ module Excon
         datum[:headers]['Host']   ||= datum[:host] + port_string(datum)
       end
 
+      # RFC 7230, section 5.4, states that the Host header SHOULD be the first one # to be present.
+      # Some web servers will reject the request if it comes too late, so let's hoist it to the top.
       if host = datum[:headers].delete('Host')
         datum[:headers] = { 'Host' => host }.merge(datum[:headers])
       end
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -252,6 +252,10 @@ module Excon
         datum[:headers]['Host']   ||= datum[:host] + port_string(datum)
       end
 
+      if host = datum[:headers].delete('Host')
+        datum[:headers] = { 'Host' => host }.merge(datum[:headers])
+      end
+
       # if path is empty or doesn't start with '/', insert one
       unless datum[:path][0, 1] == '/'
         datum[:path] = datum[:path].dup.insert(0, '/')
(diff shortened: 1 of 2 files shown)


geemus:
Thanks for the detailed report. Looks good.

Testing is trickier here, I think it might require a custom server to check things. Did you have any particular ideas/thoughts?

andrepiske:
Hey sorry for the delay.
It seems the `request_headers.ru` file already had enough to test this :)

Let me know if this seems good to you!

geemus:
Good catch, I forgot we already had that (it's been a while since I needed to dig deeper).

lib/excon/connection.rb
```diff
@@ -252,6 +252,10 @@ def request(params={}, &block)
         datum[:headers]['Host']   ||= datum[:host] + port_string(datum)
       end
 
+      if host = datum[:headers].delete('Host')
```
geemus, line 255: Overall this looks great, thanks for your detailed report and working through it with me. Just one small request and I think we should be good.

Could you add a comment here that notes the purpose (ensuring Host is first) and refers to the RFC # and section. I just worry that otherwise we may forget over time what the importance/intention of this is and it will be much easier to find/remember in a quick comment than to dig back through PRs and history. Does that seem reasonable to you?

geemus (changes requested):

lib/excon/connection.rb
andrepiske: Hey! Sorry once again for the long delay.
Sure thing, I will soon add the comments as you requested.

geemus: No problem, I appreciate it!

geemus (approved):

geemus:
@andrepiske Thanks!

Released as part of 0.77.0.

PR #726, opened 2020-09-29
Improve instrumentor for  logging errors

badlamer:
I added request body and headers to instrumentor for `error_call`. 

This give possibility debug response bodies, when recieve unexpected errors.

### DIFF
--- a/lib/excon/middlewares/instrumentor.rb
+++ b/lib/excon/middlewares/instrumentor.rb
@@ -12,7 +12,7 @@ module Excon
 
       def error_call(datum)
         if datum.has_key?(:instrumentor)
-          datum[:instrumentor].instrument("#{datum[:instrumentor_name]}.error", :error => datum[:error]) do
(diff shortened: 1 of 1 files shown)


geemus:
Unfortunately, I believe this would break things for existing instrumentors (https://github.com/excon/excon/tree/master/lib/excon/instrumentors), which only expect 2 parameters. That being said, I believe the response body already gets added to the error itself (at least for status errors), so I think you should be able to access it by something like `params[:error].response` within the instrumentor. Does that get you what you need?

badlamer:
> Unfortunately, I believe this would break things for existing instrumentors (https://github.com/excon/excon/tree/master/lib/excon/instrumentors), which only expect 2 parameters. That being said, I believe the response body already gets added to the error itself (at least for status errors), so I think you should be able to access it by something like `params[:error].response` within the instrumentor. Does that get you what you need?

Yes. Thank you.

I did not notice that the `params[:error]` is instance of Exception instead String.

geemus:
Sounds good, hopefully that will get you what you need, but feel free to reach out if you have other questions or suggestions.

Issue #733, opened 2021-01-07
Shindo tests hang (run indefitely) on random test


pvalena:
it hangs (livelock) on, f.e.:

```
DEBUG: + shindont
DEBUG:
DEBUG:   Excon bad server interaction +++++
DEBUG:   Excon basics +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
DEBUG:   Excon streaming basics #
DEBUG:   Excon basics (Basic Auth Pass) +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
DEBUG:   Excon basics (Basic Auth Fail) +++
```
and when the test is disabled, the previous tests fail as well:
```
DEBUG: + shindont
DEBUG:
DEBUG:   Excon bad server interaction +++++
DEBUG:   Excon basics +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
DEBUG:   Excon streaming basics #
DEBUG:   Excon basics (Basic Auth Pass) +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

Or in some previous run:
```
+ shindont
  
  Excon bad server interaction +++++  
  Excon basics +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++  
  Excon streaming basics #  
  Excon basics (Basic Auth Pass) +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++  
  Excon basics (Basic Auth Fail) +++  
  Excon basics (ssl) 
```

Note that we're running during package build for Fedora, so the environment is a bit modified.
(In case you're interested: It looks [like so](https://src.fedoraproject.org/rpms/rubygem-excon/blob/master/f/rubygem-excon.spec#_63).)

It may simply be a configuration issue, please advise.

geemus:
Thanks for the heads up and details. I'm not sure offhand on this one. We had some hangs in truffleruby reported for shindo more generally, which I think may have been fixed. Other than that nothing jumps immediately to mind (we haven't received any other similar issues that I'm aware of). Having it fail on different tests doesn't help narrow it down much either. Any ideas on how we might better pin down the issue at hand?

pvalena:
I can check which version which causes this.

[...]

PR #735, opened 2021-02-19
Do not use explicit hash for options.

voxik:
This should ensure compatibility with Rack 2.2+, which assumes that options might be passed as kwargs:

https://github.com/rack/rack/commit/871c82f0ffa269d7e2e237d3ade9a1eac26418fe

This is related to #733


geemus (approved):

geemus:
Thanks!

Issue #159, opened 2012-10-30
replace Socket.getaddrinfo with resolv usage


geemus:
May be some performance win, particularly in multi-threaded environments.

PR #739, opened 2021-03-20
Use Ruby Resolv instead of LibC for hostname lookup

carlhoerberg:
This way DNS resolver can be replaced with:

```ruby
Resolv::DefaultResolver.replace_resolvers [Resolv::DNS.new(nameserver: '8.8.8.8')]
```

Related: #159, #184

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -107,7 +107,7 @@ module Excon
         family = @data[:proxy][:family]
       end
 
-      ::Resolv.each_address(hostname) do |ip|
+      Resolv.each_address(hostname) do |ip|
         # already succeeded on previous addrinfo
         if @socket
           break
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Using this branch in production right now, polling a couple of thousand HTTPS APIs every minute or s for stats. Without it the DNS server would rate limit us and we couldn't swap it out as we had no access to modify /etc/resolve.conf.

geemus (approved):
Looks good, thanks!

rafbm:
Anyone updating to Excon v0.80.0 for this functionality, beware of #747. Some versions of Resolv cause processes to hang in an infinite loop: https://bugs.ruby-lang.org/issues/17781.

pboling:
Also: https://bugs.ruby-lang.org/issues/17748

pboling:
My detailed fix for this issue:
https://github.com/mperham/sidekiq/issues/2175#issuecomment-841026801

PR #740, opened 2021-03-20
Raise EOFError in Socket#readline instead of TypeError

carlhoerberg:
Before this patch the following error would be raised which could be
confusing:

```
Excon::Error::Socket: no implicit conversion of nil into String (TypeError)
```

Due to: `"" < nil`


geemus (approved):
Looks good, thanks!

PR #741, opened 2021-04-14
Handle Case When Proxy key exists but is set to nil

Conrimaceogain:
In testing related to using the libhoney-rb gem (via beeline-ruby) in an rspec suite, I discovered that there was a case where the `Proxy` key of `datum` is being added, but with a value of `nil`, causing the redact method to throw a nil method error. Adding this safe navigation should better handle that case.


geemus (approved):
Good catch, and thanks for the fix!

Issue #710, opened 2019-12-06
Add option to specify proxy CONNECT headers 


jjiang-stripe:
Hello! 👋 I'd like to set some custom headers on proxy CONNECT requests to be able to some log info on the proxy server based on the header. Would it be possible to add an option somewhere on the request datum to add custom headers (e.g. something like Go's http clients [ProxyConnectHeader](https://tip.golang.org/pkg/net/http/#Transport.ProxyConnectHeader))?

The CONNECT request is constructed over [here](https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb#L102), but I'm not entirely sure how `@data` is actually populated at request time. My ruby is pretty limited, but happy to help in any way I can!

geemus:
Certainly sounds pretty straightforward. Could you supply an example of what headers might look like to ensure I know what you intend before getting too far into this? Thanks!

jjiang-stripe:
Sure! I'm mainly looking to add in a basic trace header so that our proxy can log it when it receives a CONNECT, so something like `x-trace-header: some-uuid`.

geemus:
Got it, thanks. Yeah that helps, my guess was that it was something relatively simple like that, but just wanted to confirm before I potentially headed down a dead end. I think this should be fairly easy to do, but not sure how quickly I'll find time to do it.

geemus:
Oops, thought I had pinned this before.

cds2-stripe:
Thanks @geemus! We are definitely still interested in this feature.

hans-stripe:
Hey @geemus! I put together an experimental branch that adds support for these headers to both the default Excon client and any middlewares that might want to tweak the headers (which is how we'd like to use this feature internally). The hash handling is rudimentary but I wanted to put a demo together! Does this seem like a reasonable approach?

https://github.com/excon/excon/compare/master...hans-stripe:hans-ssl-proxy-headers

geemus:
@hans-stripe Hey, thanks for reaching out. Definitely seems like a step in the right direction. I think the key is probably adding the header insertion stuff in the socket, as you did.

I was wondering for the other part about the data.delete, which seemed a little weird to me. I think if you had data[:proxy][:headers] it would just work without needing to manipulate the data hash (and I think maybe manipulating it in that way might interfere with middleware). Does that make sense?

[...]

PR #742, opened 2021-04-14
Support additional headers for TLS proxies

jjiang-stripe:
Fixes https://github.com/excon/excon/issues/710

I don't write a whole lot of Ruby, so feel free to ask many questions/correct me if things look wonky! 

This adds a new `ssl_proxy_headers` option to the connection parameters that can be used to specify headers to send on the proxy CONNECT request. I added it as a separate parameter, since `@data[:proxy]` is typically used as a string value, and while users _could_ build their own `@data[:proxy]` hash, it seems a little cumbersome to use properly.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -104,6 +104,12 @@ module Excon
 
         request += "Proxy-Connection: Keep-Alive#{Excon::CR_NL}"
 
+        if @data[:proxy][:headers]
+          @data[:proxy][:headers].each do |key, value|
+            request << key.to_s << ': ' << value.to_s << Excon::CR_NL
+          end
+        end
+
         request += Excon::CR_NL
 
         # write out the proxy setup request
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -105,9 +105,7 @@ module Excon
         request += "Proxy-Connection: Keep-Alive#{Excon::CR_NL}"
 
         if @data[:ssl_proxy_headers]
-          @data[:ssl_proxy_headers].each do |key, value|
-            request << key.to_s << ': ' << value.to_s << Excon::CR_NL
-          end
+          request << Utils.headers_hash_to_s(@data[:ssl_proxy_headers])
         end
 
         request += Excon::CR_NL
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -104,8 +104,8 @@ module Excon
 
         request += "Proxy-Connection: Keep-Alive#{Excon::CR_NL}"
 
-        if @data[:proxy][:headers]
-          @data[:proxy][:headers].each do |key, value|
+        if @data[:ssl_proxy_headers]
+          @data[:ssl_proxy_headers].each do |key, value|
             request << key.to_s << ': ' << value.to_s << Excon::CR_NL
           end
         end
--- a/tests/proxy_tests.rb
+++ b/tests/proxy_tests.rb
@@ -21,7 +21,7 @@ Shindo.tests('Excon proxy support') do
       connection = nil
 
       tests('connection.data[:proxy][:host]').returns('myproxy.net') do
-        connection = Excon.new('http://foo.com', :proxy => 'https://myproxy.net:8080', :ssl_proxy_headers => {'x-proxy-id': 'abc123' })
+        connection = Excon.new('http://foo.com', :proxy => 'https://myproxy.net:8080')
         connection.data[:proxy][:host]
       end
 
@@ -31,10 +31,6 @@ Shindo.tests('Excon proxy support') do
 
       tests('connection.data[:proxy][:scheme]').returns('https') do
         connection.data[:proxy][:scheme]
-      end
-
-      tests('connection.data[:proxy][:headers]').returns({ 'x-proxy-id': 'abc123' }) do
-        connection.data[:proxy][:headers]
       end
     end
 
@@ -92,7 +88,7 @@ Shindo.tests('Excon proxy support') do
         connection = nil
 
         tests('connection.data[:proxy][:host]').returns('mysecureproxy') do
-          connection = Excon.new('https://secret.com')
+          connection = Excon.new('https://secret.com', :ssl_proxy_headers => {'x-proxy-id': 'abc123' })
           connection.data[:proxy][:host]
         end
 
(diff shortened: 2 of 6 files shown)


tests/proxy_tests.rb
```diff
@@ -100,6 +100,10 @@ def env_proxy_tests(env)
           connection.data[:proxy][:scheme]
         end
 
+        tests('connection.data[:proxy][:headers]').returns({ 'x-proxy-id': 'abc123' }) do
```
geemus, line 103: These tests are good for ensuring the data mutation into proxy happens as expected (unless we decide not to do that).

I think it would be great to also add a test toward the end of this file, where it uses the `proxy.ru` rackup. These tests actually run a server, which receives proxy stuff and echos it back (so you can be more confident it was sent as expected). I think it would be great to also test the roundtrip on these proxy settings, though this would require updating `proxy.ru` as well. I think maybe we could update it so the response passes back the value of `x-proxy-id` and check the response headers to gain this confidence.

Does that make sense? What do you think?

geemus (changes requested):
@jjiang-stripe Thanks for your help! I like the separate key in the config, as I definitely think that seems cleaner/safer. I've added some areas for discussion and possible refinement, and  don't think we are too far off from a great addition. Thanks!

lib/excon/ssl_socket.rb
```diff
@@ -104,6 +104,12 @@ def initialize(data = {})
 
         request += "Proxy-Connection: Keep-Alive#{Excon::CR_NL}"
 
+        if @data[:proxy][:headers]
```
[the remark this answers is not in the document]
jjiang-stripe: ah yes that makes total sense! I'll update in a bit

tests/proxy_tests.rb
jjiang-stripe: yep! always happy to write more tests 😁 I'll dig into that

jjiang-stripe:
okay updated the @data parameters and also moved the header hash to string parsing into a helper function!

I wasn't able to get the test set up properly working with the proxy CONNECT request though. it seems like Sinatra does not like CONNECT requests :( I pushed my (failing) wip tests in case you have any ideas on how I can fix them.

geemus:
@jjiang-stripe Looks great overall, and I appreciate you extracting the shared code to utils for the header writing.

I'll have to research proxy testing more, I recall us having some challenges there before, and don't know the answer off hand.

jjiang-stripe:
So I've found that adding `"--env", "production"` to [this line ](https://github.com/excon/excon/blob/e29b313b18147f4943db86fb9ff4b735f7f5cca7/tests/test_helper.rb#L248) makes the proxy CONNECT succeed (it was erroring with `env missing required key SERVER_NAME`, so seems like we were missing some env variables for rack?)

Then we end up with an SSL error `SSL_connect returned=1 errno=0 state=error: wrong version number (OpenSSL::SSL::SSLError) (Excon::Error::Socket)` since we're not actually proxying anything with our mock proxy (nor do we really want to write a full http connect proxy here 😅 ). I'm not sure if there's really a way to get around this unfortunately. I've reverted my test commit for now, if you're open to just merging this without tests 😬

jjiang-stripe:
oof just uncovered an unexpected consequence of using `[:ssl_proxy_headers]` vs `[:proxy][:headers]`. my actual use case is setting `[:ssl_proxy_headers]` in a middleware, so it's only set on the `datum` and not on `@data` by the time we initialize the socket [here](https://github.com/excon/excon/blob/1e5a5eb912db732e1d545b7b17f4cf621b95a27a/lib/excon/connection.rb#L118). 

It was working when we were using `[:proxy][:headers]` since the `[:proxy]` hash was shared between `datum` and `@data` due to the `@data.merge` over [here](https://github.com/excon/excon/blob/1e5a5eb912db732e1d545b7b17f4cf621b95a27a/lib/excon/connection.rb#L224). since `[:ssl_proxy_headers]` is used in the ssl socket's `initialize`, the `socket.data = datum` line [here](https://github.com/excon/excon/blob/1e5a5eb912db732e1d545b7b17f4cf621b95a27a/lib/excon/connection.rb#L118) doesn't end up writing `[:ssl_proxy_headers]` to the socket's `@data` until after it's already been initialized.

there's a couple options here, but i'm not sure what's the best:
1. go back to using `[:proxy][:headers]`
2. make the [socket function](https://github.com/excon/excon/blob/1e5a5eb912db732e1d545b7b17f4cf621b95a27a/lib/excon/connection.rb#L456-L465) take `datum` as a parameter instead of using `@data` (which feels like maybe the original intent of the `socket.data = datum` line?)
3. explicitly overwrite `@data[:ssl_proxy_headers] = datum[:ssl_proxy_headers]` before intializing the socket

what do you think?

geemus:
@jjiang-stripe thanks for the ongoing work here.

For the time being, I think having it with fewer tests is probably ok. I think I see some possibility for using the webrick proxy server class to actually test all of this, but I don't think we need to block on that, as I suspect it will be a little while before I get to it.

I hadn't considered the data vs datum issue, so thanks for finding that and considering the options. I'm not sure we've had a case previously where middlewares impacted socket settings, so it may have just not come up before.

From those options, I think adding an argument to socket which takes a datum sounds like a good approach to solve this issue and also make things a little more consistent in usage and expectations. Does that sound good to you?

jjiang-stripe:
I was leaning towards the socket argument option too! I just updated the code. I'm a little confused why it appears we're initializing a new socket for each `socket.write` call (maybe I'm reading it wrong? ruby confuses me 😅), but I kept it the same to be safe and passed the `datum` into each call.

geemus:
I believe it is reusing sockets, but it's easy to miss. The socket method has an `||=` in it which should either use the existing socket, or set one if it doesn't yet exist. Because of that, unless the socket_key changes it should reuse the socket (at least within that thread).

geemus:
Also, fwiw, here is a quick issue I just created around improved proxy testing with some of the more promising candidates: https://github.com/excon/excon/issues/746

I hope to return to it at some point, but haven't had a lot of time for development lately, so we'll see.

geemus:
@jjiang-stripe Thanks! Just let me know if you have other questions or suggestions. I'll try to get a release out soon (maybe early in the week, as I prefer not to do it on the weekend when I'd be less likely to notice if something goes wrong with it).

jjiang-stripe:
oh nice I see that bit for reusing sockets now. thanks for all your help here! 😁

geemus:
Released in v0.81.0.

PR #745, opened 2021-04-23
allow to use Syslog::Logger

Fivell:
https://ruby-doc.org/stdlib-2.6.3/libdoc/syslog/rdoc/Syslog/Logger.html has no `log` method defined, but `add` defined
https://ruby-doc.org/stdlib-2.6.3/libdoc/logger/rdoc/Logger.html has `log` method as alias for `add`

both loggers  have shortcut `info`  method

### DIFF
--- a/lib/excon/instrumentors/logging_instrumentor.rb
+++ b/lib/excon/instrumentors/logging_instrumentor.rb
@@ -40,7 +40,7 @@ module Excon
         end
       end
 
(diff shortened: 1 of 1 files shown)


geemus:
Looks good, thanks!

I'll probably try for a release early next week (don't like to do it on the weekend when I'd be less likely to notice if something went wrong).

geemus:
Released in v0.81.0.

PR #752, opened 2021-06-26
Add ssl_security_level support

rwg-stripe:
This change adds an `:ssl_security_level` option to change the security level of the `SSLContext` before Excon uses it to create an `SSLSocket`.

OpenSSL 1.1.0 introduced security levels, where increasingly large security levels forbid increasingly strong cryptographic algorithms and parameters.  Security level 0 forbids nothing (the default behavior in previous versions of OpenSSL), and security level 5 allows only very strong algorithms and very long key lengths.

Every OpenSSL library now has a default security level compiled into it, and programs may need to change the security level from its default.

Changing an `OpenSSL::SSL::SSLContext`'s security level requires Ruby 2.4.0+ or version 2.0.0+ of the `openssl` gem.  See the following for more information:

- https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-security_level-3D
- https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_security_level.html

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -99,6 +99,7 @@ module Excon
     :ssl_version,
     :ssl_min_version,
     :ssl_max_version,
+    :ssl_security_level,
     :ssl_proxy_headers,
     :ssl_uri_schemes,
     :tcp_nodelay,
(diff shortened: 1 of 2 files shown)


geemus (approved):
Looks good, thanks for the detailed report and patch!

geemus:
Released in v0.83.0.

PR #759, opened 2021-09-02
[fix] logging behavior for query hashes

criess:
* we must log only the query when testing for the query as hash
  * see modified behavior inside Excon::Instrumentors::LoggingInstrumentor
  * added very simple test

We found out when our logs got flooded with excon client options information, so I decided to fix it right away for upstream projects.

Let me know when you need changes, other ideas on this PR. Anyways patch is very minor.

Thanks, CR

### DIFF
--- a/tests/instrumentors/logging_instrumentor_tests.rb
+++ b/tests/instrumentors/logging_instrumentor_tests.rb
@@ -26,6 +26,33 @@ Shindo.tests('logging instrumentor') do
     ]
   end
 
+  tests("connection logger with query as hash").returns(true) do
+    Excon.stub({:method => :get}, {body: 'body'}, status: 200})
+
+    log_path = "/tmp/excon_#{Time.now.to_i}.txt"
+    logger = Logger.new(log_path)
+    # omit datetime to simplify test matcher
+    logger.formatter = proc do |severity, datetime, progname, msg|
+      "#{msg}\n"
+    end
+
+    connection = Excon.new(
+      'http://127.0.0.1:9292',
+      instrumentor: Excon::LoggingInstrumentor,
+      logger: logger,
+      mock: true
+    )
+    response = connection.request(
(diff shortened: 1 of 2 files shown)


tests/instrumentors/logging_instrumentor_tests.rb
```diff
@@ -26,6 +26,33 @@
     ]
   end
 
+  tests("connection logger with query as hash").returns(true) do
+    Excon.stub({:method => :get}, {body: 'body'}, status: 200})
```
geemus, line 30: It appears there is an extra `}` after the status portion here that is causing a syntax error in tests. Could you fix it please?

geemus (changes requested):
Looks good overall. This was almost assuredly the original intention of the code, it's interesting that it was overlooked for so long. Thanks for working on and sharing a fix!

tests/instrumentors/logging_instrumentor_tests.rb
```diff
@@ -26,6 +26,33 @@
     ]
   end
 
+  tests("connection logger with query as hash").returns(true) do
```
geemus, line 29: Hey, it appears that this test is returning false (instead of true) in CI and failing. Could you dig in and see if you could get that to passing? Thanks!

geemus (changes requested):

tests/instrumentors/logging_instrumentor_tests.rb
criess: The logfile name can be random so i used Tempfile object create example logfiles that can be appended to. Should work now …

criess: commit: bb90e2c should have the patch

geemus (approved):
Looks good. Thanks again for the report and then working through the troublesome test bits!

PR #766, opened 2021-11-02
A few patches from Debian

terceiro:
Hi, 

I was reviewing the excon package in Debian and decided to forward these patches. It would be very useful to reduce our delta; while these changes are not necessary for you, they should be harmless and make some parts of the code "more correct" is that makes sense. 

It will be easier to review this commit by commit. The commit messages should provide enough context for each change.


geemus (approved):
Yep, makes sense, just hadn't come up previously. Thanks!

Issue #473, opened 2015-02-10
Invalid SSL certificate file.


betam4x:
When using Excon I get the following: 

    Excon::Errors::SocketError: Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, `ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, `ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback), or `Excon.defaults[:ssl_verify_peer] = false` (less secure).

Working on a Windows 10 machine, but i've reproduced in a couple Windows 7 environments, haven't tried a mac or linux yet.  the SSL_CERT_FILE is set properly and mucking with the defaults for Excon does nothing.  Only disabling verification altogether works.  Also, other libraries I've used (that use OpenSSL) work fine, so it appears to be specific to Excon.

    irb(main):004:0> puts ENV['SSL_CERT_FILE']
    C:\Ruby21-x64\cacert.pem

geemus:
Were other libraries you used also in Ruby or other stuff? Could you point out specific ones? I might be able to spelunk in their code to figure out what might be different. Otherwise I'm afraid I don't have much experience on the windows side of things.

johngtimms:
Hi. I am experiencing the same issue on Windows 7, so hopefully I can provide some specifics. I have a team of people working on a Rails project and we just set up environments so this issue is very fresh in my memory. 

When you install Ruby 2.1 on Windows (using [RubyInstaller](http://rubyinstaller.org/)) and try to install gems (say, like `gem install bundler`) you will get an SSL error. I don't have the exact error in front of me, but following the steps [here](https://gist.github.com/fnichol/867550) fixes the problem and you can then use `gem`. Notice that the solution involves setting the `SSL_CERT_FILE` environment variable.

The Excon error that everyone gets says to set that same environment variable, but most people on Windows already have that variable set (assuming they can install gems!). I can confirm the report from @betam4x :

    Microsoft Windows [Version 6.1.7601]
    Copyright (c) 2009 Microsoft Corporation.  All rights reserved.
    
    C:\Users\John>irb
    DL is deprecated, please use Fiddle
    irb(main):001:0> puts ENV["SSL_CERT_FILE"]
    C:\Ruby\ca-bundle.crt
    => nil
[...]

PR #770, opened 2022-01-11
Use system OpenSSL cert file and dir if available

stanhu:
For users that manage their own SSL certs, it is a surprise that excon
ships its own root certificates. As a result, using an old version of
excon can cause SSL failures with third-party services, such as S3
(https://github.com/excon/excon/issues/473).

We now fall back to the OpenSSL defaults if available to avoid depending
on the shipped `cacert.pem`.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -39,15 +39,14 @@ module Excon
         ssl_context.max_version = @data[:ssl_max_version]
       end
 
-
       if @data[:ssl_verify_peer]
         # turn verification on
         ssl_context.verify_mode = OpenSSL::SSL::VERIFY_PEER
 
-        if (ca_file = @data[:ssl_ca_file] || ENV['SSL_CERT_FILE'])
+        if (ca_file = @data[:ssl_ca_file] || ENV['SSL_CERT_FILE'] || use_if_readable(OpenSSL::X509::DEFAULT_CERT_FILE))
           ssl_context.ca_file = ca_file
         end
-        if (ca_path = @data[:ssl_ca_path] || ENV['SSL_CERT_DIR'])
+        if (ca_path = @data[:ssl_ca_path] || ENV['SSL_CERT_DIR'] || use_if_readable(OpenSSL::X509::DEFAULT_CERT_DIR))
           ssl_context.ca_path = ca_path
         end
         if (cert_store = @data[:ssl_cert_store])
@@ -190,5 +189,10 @@ module Excon
       @data[:client_key_pass] || @data[:private_key_pass]
     end
 
+    def use_if_readable(filename)
+      return filename if File.readable?(filename)
+
+      nil
+    end
   end
 end


stanhu:
Tests will be fixed with https://github.com/excon/excon/pull/771.

stanhu:
Hmm, maybe the `set_default_paths` handles this already. I wonder if the `add_file` call in https://github.com/excon/excon/pull/770/files#diff-40a9ad671d331178cbbc470b7eabb2455ba0e1c52df329852fb6f4780c9c72f5R58-L67 may cause issues in validation. With the recent LetsEncrypt root expiry, removal of the expired cert was important for validation to work properly.

geemus:
@stanhu Hey, apologies for any surprise/confusion around this. I'll do my best to share my recollection of the status of this and hopefully we can get to the bottom of it.

My understanding was that the `set_default_paths` call would establish a path that included the openssl cert. And the bundled cert would then be added at the end of the path, which would mean that it should only ever be used if the default paths do not include one already.

So my expectation is that it would use OpenSSL instead of the bundled if it exists. If I recall correctly we set it up this way because there were certain environments/installations where OpenSSL certs were not present and we wanted it to still work (I want to say this was a Windows thing, but it's been a long time now). It definitely wasn't supposed to be instead of OpenSSL, but just to make sure that it would "just work" in other cases.

So, if that isn't the behavior we are actually seeing, we should change it.

I suspect that because of the way we are doing path and the particular case (removal of an expired), it might mean that after failing to be found in the OpenSSL group it would try the next thing in the path (the bundled) which might then fail? Does that make sense and/or sound likely? Maybe we should change it to check in some way if `set_default_paths` added something (in which case no action is needed), vs nothing being added and needing to add the bundled one in. I think that would still allow things to "just work" in the original use case while avoiding the surprising edge case you ran into. What do you think?

stanhu:
@geemus Makes sense. I think I'll close this. I think it threw me off that `set_default_paths` is called, and then potentially an *old* list of certs get added to it. 

In https://gitlab.com/gitlab-org/gitlab/-/issues/342326#note_694301994, we saw that if you called `set_default_paths` and called `add_file` without parsing individual certs, validation wouldn't work quite right, at least when it came to the LetsEncrypt root certificate.

https://gitlab.com/gitlab-org/gitlab/-/merge_requests/71697/diffs fixed the problem for us.

I suspect the `add_file` isn't doing what you think it is doing, and perhaps only adding the first certificate in `cacert.pem`.

geemus:
@stanhu Got it, thanks for the update. We could certainly add better documentation/explanation around this for our future selves and others. Otherwise I'm certainly open to improving the usage here if we need to. I looked over the issue/diffs you mentioned but fear I don't have enough context/memory of things to immediately grok what we would need to change. So let me know if you think action is required and if you have advice and I can try to find some time to dig in. Otherwise it sounds like maybe it wasn't as much of a problem as you originally expected and maybe we can let it be for now?

PR #774, opened 2022-01-22
Fix debug instruction in README

grobie:
The README seems to have gotten out-of-sync with reality, the options
`:debug_request` and `:debug_response` don't have any effect anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Here are a few common examples:
 
 ```ruby
 # Output debug info, similar to ENV['EXCON_DEBUG']
-connection = Excon.new('http://geemus.com/', :debug_request => true, :debug_response => true)
+connection = Excon.new('http://geemus.com/', :debug => true)
 
 # Custom headers
 Excon.get('http://geemus.com', :headers => {'Authorization' => 'Basic 0123456789ABCDEF'})


geemus:
@grobie - Unfortunately, I think it's not quite so clear cut. These options do something, just conditionally. Though you are certainly correct that the usage documented in the README appears to be inaccurate.

In particular, I think it does have an impact in two situations:
1. When errors are reported, it determines what information is included.
2. When the `StandardInstrumentor` is attached, it determines what information is included.

The `debug` option itself actually just sets these two options together, but it also attaches the `StandardIntrumentor`. I would argue that this is what is missing for the two options you call out when they are used by themselves. You can see the behavior for the debug option here:

https://github.com/excon/excon/blob/5f384be869097fb5195ac086e8abb714e247b570/lib/excon/connection.rb#L86-L89

I think we might want to change this, or add to it. So that if any of `debug`, `debug_request`, or `debug_response` are set, the StandardInstrumentor will be attached. Does that all make sense? What do you think?

grobie:
That makes sense to me as additional change. To be honest, it was just the third time where I needed a quick way to debug a request and followed the README without success.

Keeping the README example simple appears to be desirable to me, especially as it's the first example and says "similar to ENV['EXCON_DEBUG']". Changing the behavior of debug_request/debug_response on top sounds like another PR?

geemus:
@grobie Thanks for the quick response. I agree that keeping it simple sounds good, that having the README be more accurate is important, and that if/when we make those other changes they would make sense to happen in a distinct PR. I guess I'm getting a bit too wrapped up in the details. I think this change as is makes sense, thanks for the fix and talking through it!

geemus (approved):

geemus:
NOTE: The test failure is unrelated (it has to do with ruby/rubygems versions and github actions). It's subsequently be fixed on master, but as this is just a documentation change I'm going to go ahead and merge rather than worrying about rebasing or similar.

grobie:
@geemus I'm a bit bored (in a quarantine for a week now). If you want someone else to take a look at your other suggestion, I'm up for it.

geemus:
@grobie Sorry to hear about the quarantine (and boredom). If you'd like to work on that refinement for the request/response stuff that sound great to me. Hopefully it was relatively clear from above what we could do to align what I think the intention was with what actually happens, but don't hesitate to circle back if you get stuck or have questions. Thanks!

PR #778, opened 2022-01-23
add GET as default method for Connection#request

### LINKED ISSUE #775: Unexpected Connection#request interface
The `Connetion#request` method does neither enforce the existence of the `:method` parameter, nor does it set a default. The result is that `Excon.new("https://www.google.com").request` creates and sends an invalid HTTP request missing an HTTP verb.

Given `#request` uses `/` as default path if none is given, I'd expect the method to default to :get. An exception would be reasonable as well to require setting the method explicitly.

geemus:
closes #775

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -253,6 +253,11 @@ module Excon
         datum[:headers] = { 'Host' => host }.merge(datum[:headers])
       end
 
+      # default to GET if no method specified
+      unless datum[:method]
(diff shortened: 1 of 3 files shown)


grobie:
Thanks a lot for the quick fix!

geemus:
Sure thing. I don't think I had personally stumbled across it or heard others mention it, but it made good sense when you mentioned it and was easy enough to change. Thanks for the idea and support!

PR #779, opened 2022-01-23
Use StandardInstrumentor if any debug mode is enabled

grobie:
So far the StandardInstrumentor was only set for the :debug helper
setting, but not when the specific :debug_request or :debug_response
settings were used.

---

@geemus Well, I feel silly for opening such a simple PR :sweat_smile:. I don't immediately see how to test this as I'm unfamiliar with the code base and internals, so this is literally just how I understood your suggestion.

Besides the tests, my biggest question is around the behavior when the user has already set the `:instrumentor` option? Does it need some dispatch instrumentor which can call multiple instrumentors at the same time? I haven't seen anything like that immediately.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -79,12 +79,11 @@ module Excon
 
       setup_proxy
 
-      if ENV.has_key?('EXCON_STANDARD_INSTRUMENTOR')
-        @data[:instrumentor] = Excon::StandardInstrumentor
-      end
-
       if @data[:debug] || ENV.has_key?('EXCON_DEBUG')
         @data[:debug_request] = @data[:debug_response] = true
+      end
+
+      if @data[:debug_request] || @data[:debug_response] || ENV.has_key?('EXCON_STANDARD_INSTRUMENTOR')
         @data[:instrumentor] = Excon::StandardInstrumentor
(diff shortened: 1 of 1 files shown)


geemus:
No worries, it was good to talk through, even if the change itself was somewhat trivial.

You are right that this could override an instrumentor that was otherwise set, but I think that would be expected/intended (you want to override stuff and get to debug output for some reason). The other options already worked that way, and these were at least intended to provide more nuance to support that, so I think they are still in the same spirit. Maybe we should think about more flexibility here, but it hasn't come up as needed so far.

Now I find myself second guessing this though as I again review the code. (apologies, it's been some time since I'd looked at this part). It looks like I was accurate in saying it changes the contents of error messages, and maybe that is actually good enough/expected after all.

The earlier version of the README was clearly wrong, which you have now fixed. And upon further reflection I think maybe we should just leave this be (at least for now) and have it just effect error handling. Apologies for the confusion, I think I confused myself as well (we've been without childcare due to pandemic related reasons for over a week now, so things have been a little hectic). I hope you'll forgive the confusion, especially given that it was at least too large of a change to figure out.

Just let me know if you disagree, have questions, or would like to discuss this further. Thanks!

PR #793, opened 2022-09-29
Use `ssl_verify_peer_host` for SNI if configured

davidtaylorhq:
When set, `ssl_verify_peer_host` is used to verify the certificate of the SSL connection after it's opened.

https://github.com/excon/excon/blob/226924e59db71a1c8ce54a682cb343c55be44219/lib/excon/ssl_socket.rb#L155-L158

To ensure the server responds with the appropriate certificate, we need to use this same `ssl_verify_peer_host` value for SNI as well.

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -133,7 +133,7 @@ module Excon
 
       # Server Name Indication (SNI) RFC 3546
       if @socket.respond_to?(:hostname=)
-        @socket.hostname = @data[:host]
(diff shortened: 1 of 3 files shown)


geemus:
@davidtaylorhq Do you need a release with this fix or is it ok to wait for the next release? (usually updated default certs is the forcing function on that, which tends to happen every month or two).

davidtaylorhq:
Thanks for the quick merge @geemus! If you're able to cut a release without too much effort, that would be great. Otherwise, I'll source the gem from github in the meantime.

geemus:
Sounds good, just released as part of v0.93.0.

PR #798, opened 2022-11-11
Remove from list of contributors

sandstrom:
It's an old email address, but easiest to just remove.

Thanks for an awesome gem! 💎

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,6 +1,5 @@
 * Aaron Stone <aaron@serendipity.cx>
 * Adam Esterline <adam@esterlines.com>
-* Alexander Sandström <alexander@skovik.com>
 * Andrew Katz <andrew.katz@outright.com>
 * Andy Delcambre <adelcambre@gmail.com>
 * Anshul Khandelwal <anshul@anshulkhandelwal.com>


geemus:
Hey, thanks for the fix and glad you like it! Historically, I've used a script to regenerate this file based on the emails that appear in the git commit history. I'm a bit forgetful about doing this regularly (probably should integrate/automate it). The downside to that is that even if we make the change you propose here, I believe the old email will reappear the next time I run this script (because that old email is still in the history). Could you help me understand your concerns about the old email appearing here? I might need to rethink how we do this more broadly. Thanks!

sandstrom:
@geemus No major concerns really, just wanted to remove the email for spam reasons (not sure if it makes any difference though).

One approach could be to generate the list without emails (names only). I know one can still dig through the git history and find emails, but that's less of a concern to me.

I'm also fine with keeping this as-is if you think it's a hassle.

geemus:
Yeah, that makes perfect sense. I think names without emails sounds like a good place to land this.

The actual generation/etc happens with https://github.com/geemus/osrcry, fwiw. I'll see about updating that and then regenerating this some time soonish (and then I can close this issue out as well.

Thanks for the feedback and ideas!

PR #803, opened 2022-12-20
Add scheme, query and method attributes to the response

### LINKED ISSUE #795: Include request method in response
When using instrumenters to set up logging or tracing the `excon.response` notification is particular helpful as it includes the response status and request latency. The notification data already includes information about request host, port, and path among others, but notably the request method is missing.

Especially in RESTful API design it's common that different operations share the same request path. The absence of the method attribute makes it difficult to disambiguate response notifications.

Related discussion https://github.com/excon/excon/issues/757.

mburumaxwell:
Adds `scheme`, `query`, and `method` attribute to the response. This makes it easier for logs and when writing URIs for responses using `Excon::Utils.request_uri(response)`

Fixes #795

### DIFF
--- a/lib/excon/response.rb
+++ b/lib/excon/response.rb
@@ -32,6 +32,9 @@ module Excon
     def path
       @data[:path]
     end
+    def query
+      @data[:query]
+    end
     def port
       @data[:port]
     end
@@ -78,6 +81,7 @@ module Excon
         :scheme          => datum[:scheme],
         :headers       => Excon::Headers.new,
         :path          => datum[:path],
+        :query          => datum[:query],
         :port          => datum[:port],
         :status        => status,
         :status_line   => line,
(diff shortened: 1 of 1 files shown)


mburumaxwell:
@geemus seems the CI/test against the latest Ruby fails even in the past actions. Is there anything, in particular, to do in this PR?

geemus:
@mburumaxwell thanks for checking in. You are correct, there is something in head that has caused failures for some time. I've kept it in the CI mix to keep an eye on it, but it isn't a blocker. As I'd mentioned elsewhere, we may want eventually to add some tests around this, but I think it looks fine for now. Apologies for the delay as I was traveling, I'll try to give this one last review and get it in shortly.

geemus (approved):
LGTM

geemus:
Thanks!

mburumaxwell:
Awesome

Issue #805, opened 2022-12-30
Change to Ruby URI gem breaks unix:// URIs


jeffparsons:
This change makes the URI `host` field an empty string instead of `nil` when absent: https://github.com/ruby/uri/commit/81263c9e94bd67ca01deee238842a88c2c8885f3

This line in excon appears to assume it will be nil: https://github.com/excon/excon/blob/efd48747fe6c6fa959e787aa5949241cd762f8f3/lib/excon/connection.rb#L92

Discovered via 'docker-api' gem. Example IRB session:

```
irb(main):001:0> require 'docker-api'
=> true
irb(main):002:0> Docker.url = 'unix:///var/run/docker.sock'
=> "unix:///var/run/docker.sock"
irb(main):003:0> Docker::Container::all
/home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/excon-0.95.0/lib/excon/connection.rb:93:in `initialize': The `:host` parameter should not be set for `unix://` connections. (ArgumentError)                     
When supplying a `unix://` URI, it should start with `unix:/` or `unix:///`.                    
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/excon-0.95.0/lib/excon.rb:142:in `new'
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/excon-0.95.0/lib/excon.rb:142:in `new'
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/docker-api-2.2.0/lib/docker/connection.rb:35:in `resource'                                                                                         
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/docker-api-2.2.0/lib/docker/connection.rb:45:in `request'                                                                                          
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/docker-api-2.2.0/lib/docker/connection.rb:109:in `block (2 levels) in <class:Connection>'                                                          
        from /home/jeff/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/docker-api-2.2.0/lib/docker/container.rb:343:in `all'                                                                  
        from (irb):3:in `<main>'                                    
[...]

PR #806, opened 2022-12-31
Convert empty host string from 'uri' to `nil`

jeffparsons:
'uri' gem >= v0.12.0 (which the version of the default gem in Ruby 3.2) returns an empty string instead of nil for no host. Convert it back to `nil` as soon as possible so the rest of the code doesn't have to care about handling different versions of 'uri'.

Fixes https://github.com/excon/excon/issues/805.

### DIFF
--- a/lib/excon.rb
+++ b/lib/excon.rb
@@ -126,7 +126,10 @@ module Excon
         raise ArgumentError.new("Invalid URI: #{uri}")
       end
       params = {
-        :host       => uri.host,
+        # 'uri' >= v0.12.0 returns an empty string instead of nil for no host.
+        # Convert it back here so the rest of the code doesn't have to care about
+        # handling different versions of 'uri'.
+        :host       => (uri.host && !uri.host.empty?) ? uri.host : nil,
         :hostname   => uri.hostname,
         :path       => uri.path,
         :port       => uri.port,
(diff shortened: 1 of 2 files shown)


jeffparsons:
BTW, @geemus, if you'd prefer a different approach, more thorough testing, etc., please don't hesitate to say! (I'm not going to be offended if you tell me everything about this is wrong 😛 and don't want you to feel you have to fix it yourself.)

geemus:
@jeffparsons thanks for the detailed report and work here. Apologies it took me a few days to have a chance to circle back, it was a hectic weekend. I do wonder if we might want to make the exceptional code fix the old way (presuming the new way will stick and eventually become the only way), rather than sort of fixing the new way, if that makes sense.

Also, looking at the code a bit, I suspect that it wouldn't have a problem with an empty string in host (instead of nil), and that the check around the error might just be too strict. Could we tweak https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L92 to address this problem to something like: `if @data[:host] && !@data[:host].empty?`? From a brief look at the code I think that would allow either case to work and would involve change in fewer places. What do you think?

geemus:
@jeffparsons - Thanks, yes that looks like what I was thinking. I believe I prefer that (at least at the moment). I'll try to get that merged and released soon (and will close this in the mean time, before I forget).

PR #807, opened 2023-01-03
Treat 'host' in URI as present only if non-empty

jeffparsons:
'uri' >= v0.12.0 returns an empty string instead of nil for no host. So treat the parameter as present if and only if it is both non-nill and non-empty.

This is an alternative to https://github.com/excon/excon/pull/806.


geemus (approved):
LGTM, thanks!

geemus:
Fixes #805
(this was noted in #806, but didn't get ported it looks like).

PR #808, opened 2023-01-09
Convert proxy host value to nil if needed.

voxik:
`URI.parse` used to return `nil` for `host`, when parsing `unix:` URI. However, it returens empty string since Ruby 3.2 / URI v0.12.0 for security reasons:

https://hackerone.com/reports/156615
https://github.com/ruby/uri/commit/81263c9e94bd67ca01deee238842a88c2c8885f3

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -579,6 +579,8 @@ module Excon
           raise ArgumentError, "The `:ssl_proxy_headers` parameter should only be used with HTTPS requests."
         end
         if @data[:proxy][:scheme] == UNIX
+          # URI.parse might return empty string for security reasons.
+          @data[:proxy][:host] = nil if @data[:proxy][:host] == ""
           if @data[:proxy][:host]
             raise ArgumentError, "The `:host` parameter should not be set for `unix://` proxies.\n" +
                                  "When supplying a `unix://` URI, it should start with `unix:/` or `unix:///`."


voxik:
Please note that while this test the test failure, I have no idea if it just does not re-introduce the issue which is the upstream change trying to avoid. However, the original issue is about `uri.to_s` containing the original possible malicious URL which should not be the case here, at leas not immediately.

voxik:
This is related to #807 apparently

geemus:
Thanks for the detailed report and fix.

I think since we are limiting this to only treat the empty/nil host specially if the the scheme is UNIX should protect us (I hope) from people smuggling in malicious stuff, or at least the kind of malicious stuff they mention with examples like `http:////malware.com/real/path`.

Yes, #807 was the fix for the non-proxy portion of things. And I think it should be safe-ish for the same reason (it only applies this logic if the scheme is UNIX.

I would also guess that folks are generally not allowing their users to pass in unix socket info that they then throw into excon, which should also mitigate the concern.

geemus:
@voxik good call on 3.2, I've added it here: https://github.com/excon/excon/commits/master. I used to not ignore head, but it had been consistently failing (maybe because of the issues we just fixed, as the tests are passing again). I'll try to keep an eye on it and maybe remove it from the ignore list if it keeps working well.

geemus:
Fix released in v0.97.0. Thanks again!

PR #814, opened 2023-02-02
Automatic Fork Safety

casperisfine:
Forking servers are a popular deployment method in Ruby. One big footgun is that uppon fork, all file descriptors are inherited, so if you create some FDs before fork and hold on them you may end up with the same connection being used across many process creating havoc.

As such it's a good idea for libraries that do keep persistent connections do check the PID before re-using them.

The alternative is for libraries to provide some kind of callback that the application has to call before fork, but that's easilly missed by users and the PID check is very cheap.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -62,6 +62,7 @@ module Excon
     # @option params [Class] :instrumentor Responds to #instrument as in ActiveSupport::Notifications
     # @option params [String] :instrumentor_name Name prefix for #instrument events.  Defaults to 'excon'
     def initialize(params = {})
+      @pid = Process.pid
       @data = Excon.defaults.dup
       # merge does not deep-dup, so make sure headers is not the original
       @data[:headers] = @data[:headers].dup
@@ -479,6 +480,11 @@ module Excon
     def sockets
       @_excon_sockets ||= {}
       @_excon_sockets.compare_by_identity
+
+      if @pid != Process.pid
+        @_excon_sockets.clear # GC will take care of closing sockets
+        @pid = Process.pid
+      end
 
       if @data[:thread_safe_sockets]
         # In a multi-threaded world, if the same connection is used by multiple
--- a/tests/connection_tests.rb
+++ b/tests/connection_tests.rb
@@ -36,6 +36,22 @@ Shindo.tests('Excon Connection') do
         response = connection.request(path: '/bar', method: 'get')
         response.body == 'bar'
       end
+
+      if ::Process.respond_to?(:fork)
+        connection_id = connection.send(:socket).object_id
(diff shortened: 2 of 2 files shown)


geemus:
@byroot Thanks. I think the code looks good and it makes sense, my main concern would be around performance since it adds those regular PID lookups to pretty much everything. I was trying to give myself broader context and saw your comments about performance concerns also here: https://bugs.ruby-lang.org/issues/17795. Does that remain a concern for something like this? Should we consider, for instance, wrapping this in an option so you could choose whether or not to impart this cost? (presumably somewhat similar to how we wrap the threadsafe check and logic). Does that make sense?

Orthogonally, does this mean shopify is utilizing excon in production? Not that it makes a big difference one way or the other, but it's always interesting (and rewarding) to see where it has been adopted.

byroot:
> Does that remain a concern for something like this?

Right. So at that time my concern was that I'd see `Process.pid` quite high in stackprof reports. However, since then we discovered that the way stackprof worked at the time had a tendency to inflate native methods a bit (https://github.com/tmm1/stackprof/pull/150). In the end it's really isn't as big as I feared it was.

The fork callback is still worth using IMO, but it's not always convenient. Here you'd need to keep weak references on the connections objects, which would likely be more expensive than a PID check.

Ultimately I think the PID check cost here is totally dwarfed by the time even the smallest HTTP request would take, so I'm not too concerned. 

> does this mean shopify is utilizing excon in production?

Not directly, it's pulled by one of our other dependencies (https://rubygems.org/gems/avro_turf). We kinda use all the http clients in existence because of various hard dependencies like this one `-_-`

geemus:
ah, too bad. I suspect something like "not directly" is a pretty common reason folks are using it. Similarly, having LOTS of different http clients getting looped in sounds all too familiar.

geemus (approved):
LGTM. It sounds like this should help with forks and that the PID check is fairly inexpensive. Still, we'll want to keep an eye on it and can tweak or guard it if it ends up being an issue.

byroot:
thanks for the merge!

geemus:
Also, sorry for referencing the wrong person/account instead of the one this started with, just realized when I saw multiple avatars.

geemus:
Of course, thanks for the fix. Do you need a release with the change?

byroot:
> wrong person/account

No worries, makes little difference to me.

> Do you need a release with the change?

If it's not too much bother for you that's be great. But if you have other changes planned, it can wait.

geemus:
Sounds good. I'm about to wrap up for the day and want to look at one other bug a little more. One way or the other I'll try to get a release either early tomorrow or early next week.

geemus:
Released in v0.99.0.

PR #824, opened 2023-07-04
Drop old RUBY_VERSION conditionals

voxik:
Judging by CI, it does not look that anybody would care about Ruby 1.9 and older


voxik:
The "head" error does not seem to be related to this PR

PR #827, opened 2023-07-28
Add JRuby into test matrix.

voxik:
It was previously dropped by 27d23c4 for too many failures. But since there are JRuby code paths, they should be either tested or dropped. Closing eyes does not help probably.

Please note that I am not JRuby user myself, but I care about code readability and clarity.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['2.7', '3.0', '3.1', '3.2', 'head']
+        ruby-version: ['2.7', '3.0', '3.1', '3.2', 'head', 'jruby']
     timeout-minutes: 5
 
     steps:
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   test:
-    continue-on-error: ${{ matrix.ruby-version == 'head' }}
+    continue-on-error: ${{ contains(fromJSON('["head", "jruby"]'), matrix.ruby-version) }}
     runs-on: ubuntu-latest
     strategy:
       matrix:


voxik:
Yeah, a lot of errors :/

voxik:
I've updated the PR to continue on JRuby test failures.

geemus:
Adding this back in makes sense, it's certainly unfortunate that there are so many errors. I wish we had one or more folks that were invested in it at users who could help out more.

I just restarted the tests to ensure the skips worked as expected and it seems off. In particular, I wouldn't expect jruby to be cancelled if continue on failure was working as expected. Does that make sense? I haven't tried something like that fromJson call before, so maybe something isn't quite working as expected there? Since there are just two things to check one option might be something like: `version == 'head' || version == 'jruby'`. I think that is more or less what I've done before (if memory serves). Does that make sense? It's also possible I just misremember what the behavior with continue is supposed to be like.

voxik:
> I just restarted the tests to ensure the skips worked as expected and it seems off. In particular, I wouldn't expect jruby to be cancelled if continue on failure was working as expected. Does that make sense?

I was asking the same question until I have noticed the:

https://github.com/excon/excon/blob/50566a93cf8f16eca1f4ce687f86a54869cdcf8d/.github/workflows/ruby.yml#L26

I have observed the JRuby test suite and it seems it generally runs slower. But once it begins to fail, it fails hitting some particular timeouts, until the whole test suite hits timeouts globally. But the global timeout is sensible IMHO, if somebody is going to fix those.

> I haven't tried something like that fromJson call before, so maybe something isn't quite working as expected there? Since there are just two things to check one option might be something like: `version == 'head' || version == 'jruby'`. I think that is more or less what I've done before (if memory serves). Does that make sense? It's also possible I just misremember what the behavior with continue is supposed to be like.

Being not intimately familiar with the GH actions, I was initially thinking about the `||` variant, but prior I have made sure it works, I have discovered the `fromJson` in some of the examples. I agree it is not the most readable think in the world. OTOH it is more universal if there is need to skip more versions. I can change it if you prefer (but rather waiting for confirmation ;) )

Issue #829, opened 2023-09-08
Add support to set a DNS resolver timeout


bruno-b-martins:
Hey there 👋 

I'm trying to fix an issue in a process of ours where the DNS resolver hangs for too long when resolving an internal hostname.
We still resolve the internal hostname with a retry mechanism, but the retry mechanism only kicks in after ~60 seconds.
That's basically how long it took until we receive the error message triggered by https://github.com/excon/excon/blob/master/lib/excon/socket.rb#L167:
> Excon::Error::Socket (no address for internal.hostname.com (Resolv::ResolvError))

Since we have the `Resolv::ResolvError` in there, I'm guessing that's where it hangs.
To add a shorter timeout in Resolv we need to use `Resolv::DNS#timeouts` (https://ruby-doc.org/3.2.2/stdlibs/resolv/Resolv/DNS.html#method-i-timeouts-3D).

I'll create a PR to apply that change ☝️ and I appreciate any feedback you have to give me.
Navigating through the `Connection` class, its `socket` method and all available socket classes was a bit confusing as a first time contributor 😅

geemus:
Without a reproduction case to play with, it's hard for me to know what the right fix for that particular problem, is, but I did provide some support on the linked PR. Hopefully with that you'll be able to test and see if that helps with this problem. Just let me know if you have questions or I can help further.

bruno-b-martins:
#830 just implemented this, so I'm closing the issue now.

Thanks for the help and until next time 👋

PR #830, opened 2023-09-08
Add support to set a DNS resolver timeout

bruno-b-martins:
## Goal

Be able to use a timeout in the DNS resolver step (https://github.com/excon/excon/issues/829).

## Description

This PR adds support to use `Resolv::DNS#timeouts` with a new connection and request parameter `dns_timeouts`.
It also cleans up some white spaces.

## To Do
- [x] Add specs


bruno-b-martins:
I'm struggling to find a solution that would work when a hostname is an address ('google.com') and when it is an IP address already ('127.0.0.1').

Screenshot 2023-09-09 at 10 59 16

Does anyone have an idea how to achieve that?

lib/excon/constants.rb
```diff
@@ -134,41 +135,42 @@ module WaitWritable; end
...
-    :thread_safe_sockets  => true,
-    :uri_parser           => URI,
-    :versions             => VERSIONS,
-    :write_timeout        => 60
+    :mock                => false,
+    :nonblock            => true,
+    :omit_default_port   => false,
+    :persistent          => false,
+    :read_timeout        => 60,
+    :dns_timeouts        => 60,
```
geemus, line 163: Really minor nitpick, but could you place this key alphabetically in the list? (that should be how all the existing keys are and can make it easier to find stuff in the future).

README.md
```diff
@@ -191,6 +191,10 @@ connection = Excon.new('http://geemus.com/', :connect_timeout => 360)
 
 # opt-out of nonblocking operations for performance and/or as a workaround
 connection = Excon.new('http://geemus.com/', :nonblock => false)
+
+# set shorter dns_timeouts for resolving addresses (default is 60 seconds)
```
geemus, line 195: Looking at the resolv code, I think the default is actually a bit more complicated than this (at least in so far as I understand it). It appears to be an array of values that looks like `[ 5, second = 5 * 2 / nameserver_count, 2 * second, 4 * second ]`, see: https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L342C7-L342C71. It does appear to allow for singular values, but it might be good to explain that it can be singular or an array here (and what to expect from each).

geemus:
It is definitely a bit confusing. In the usage you showed above, you are explicitly creating a DNS resolver (which is expected to return nil on hostnames, just as you expect). In our original usage, where we just use Resolv.each_address, it is using the DefaultResolver values, which it turns out is actually the array of `[Resolv::Host.new, Resolv::DNS.new]`, see: https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L87. I think we can both set the dns timeout and get this behavior, but it's a little awkward. This seemed to work for me:

```ruby
require 'resolv'

dns_resolver = Resolv::DNS.new
dns_resolver.timeouts = 60
resolver = Resolv.new([Resolv::Hosts.new, dns_resolver])

resolver.each_address('google.com') { |ip| puts ip }
# 142.250.191.174
# 2607:f8b0:4009:817::200e

resolver.each_address('127.0.0.1') { |ip| puts ip }
# 127.0.0.1
```

Does that make sense?

bruno-b-martins:
It is confusing indeed. I've also tried that out and it is "working in my machine".

Can you please confirm if I understood it right?

By passing 2 resolvers to `Resolv.new`, we're defining the resolvers that will be used in https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L118.
Which means the first resolver will be `Resolv::Hosts`, and the second one the `Resolv::DNS`.

The first one is looking into `/etc/hosts` in the machine it is running (https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L178). So, it's okay that we don't set a timeout because it shouldn't do any connection.

The second one is looking into a nameserver defined in `/etc/resolv.conf` by default (https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L311).
In this one is where we can set a timeout as there will be a connection.

Is this ☝️ right?

README.md
bruno-b-martins: I've just added some additional info.
Do you think it's enough?

geemus: Yeah, I think that looks good. It gives people an overview and a place to look for details. If we got much more into the details here, I think there would be too much risk that it would fall out of date with Resolv, so it seems better to give a description and a reference directly instead.

lib/excon/constants.rb
```diff
@@ -134,41 +135,42 @@ module WaitWritable; end
...
-    :connect_timeout      => 60,
-    :debug_request        => false,
-    :debug_response       => false,
-    :headers              => {
+    :ciphers             => 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:DES-CBC3-SHA:!DSS',
+    :connect_timeout     => 60,
+    :debug_request       => false,
+    :debug_response      => false,
+    :dns_timeouts        => 60,
```
geemus, line 145: If I remember correctly from looking through the resolv code, passing a nil timeout will let the library assign it's own default. So maybe we should just leave this default out and fallback to that? I think the default there is actually an array which depends in part on how many different places you are doing lookups to, which seems more flexible (and also it is another place we can avoid having things drift from the original implementation). What do you think?

geemus:
Implementation is looking good, I think once we figure out the right default this should hopefully be good to bring in. Thanks!

lib/excon/constants.rb
bruno-b-martins: I think that would be great!

And yes, if we pass `nil`, `Resolv` will [generate_timeouts](https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L1110) based on the nameservers count.

I'll update the default accordingly.

bruno-b-martins:
I've updated the default as you suggested.

I'm now wondering about specs. I'd like to test that `dns_timeouts` works fine, but the test's setup seems kinda complex.
I'll give it a try though and let you know if I need help.

lib/excon/constants.rb
```diff
@@ -134,41 +135,42 @@ module WaitWritable; end
...
-    :connect_timeout      => 60,
-    :debug_request        => false,
-    :debug_response       => false,
-    :headers              => {
+    :ciphers             => 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:DES-CBC3-SHA:!DSS',
+    :connect_timeout     => 60,
+    :debug_request       => false,
+    :debug_response      => false,
+    :dns_timeouts        => nil,
```
geemus, line 145: Looks good. We might want a comment here explaining that nil allows Resolv to build it's own default though (as opposed to nil meaning no timeout for instance, or something else altogether). I worry otherwise it may be confusing and/or get forgotten pretty quickly. What do you think?

geemus:
Yeah, a lot of the testing around this stuff can be a bit weird since it tends to work at a lower level than many other things expect. If you want to give it a try and then circle back, that sounds great.

lib/excon/constants.rb
bruno-b-martins: I thought that but since we had that info in the README felt like repeating ourselves.
I'll add it as I prefer to have that data closer to the code.

geemus: Sounds good, thanks!

bruno-b-martins:
I've managed to make a spec that works with a sinatra "nameserver" that timeouts.
However, I failed to make it work with the other tests...
Namely, I had to bootstrap the sinatra nameserver in a separate shell and, for some reason, it ignores the `port` (53) I defined. Which means it enters in conflict whenever one of the other sinatra servers kick in.

I tried using `include_context('test server', :webrick, 'dns_timeout.ru', before: :start, after: :stop)` like I've seen in other specs but it still was disregarding the port I've set.

I'll give it a couple more tries, but if you know already the solution to this conundrum, "I'm all ears" 😅

bruno-b-martins:
Scratch that, as I wrote the previous comment I remembered I hadn't tried the current pushed combination and it works perfectly now.

bruno-b-martins:
Actually... No. The spec is not yet finished...
It is replying way too fast, so I guess the sinatra nameserver isn't being reached. Something is failing before making the connection to the dns resolver.

I'm investigating that now.

bruno-b-martins:
@geemus , I'm struggling with the specs...

I can easily check if `dns_timeouts` was passed down to `Resolv::DNS::Config` -> first `it` example.

But I'm not managing to do a sort of integration test where we can check if setting `Resolv::DNS::Config`'s `@timeouts` (via `dns_timeouts`) stops the process if the nameserver takes longer to reply -> second `it` example.
The spec runs in under 0.3 seconds in my codespace, which indicates that it isn't waiting 1 full second until it times out...

In `Resolv`'s codebase I can see that a Timeout error is raised, but it's also [rescued](https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L1128) 😅 
There are other possible errors as well, but they're also rescued. Which makes it harder to understand what's missing on our spec. As far as I understood, these errors are meant to be used internally only. That's why it is like that.

So, the question is:
Should we try to have this kind of integration test where we test if `Resolv` times out our request? Or can we simply test if `Resolv`'s configuration is correct?

geemus:
I see the rescue you mentioned. I think that is there to support the case where you pass an array of timeouts (you don't want to error on any of the earlier failures, only if they all fail in that case I think, which I think would be the error after the each: https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L1132).

I agree that it seems harder (and more confusing) to figure out how exactly to get something to flow through the whole integration context. It looks like a few different errors (including host unreachable and connection refused) get mapped to timeout errors and either of those would resolve faster (could be the port like you mentioned). Were you able to see if the sinatra server received anything here?

As an in-between option, it looks like Resolv will also raise a timeout error if you set the timeout to 0 (or less), so we could check for that and it would get us relatively close to the integration level if this turns out to be too tricky and/or brittle to be worth it.

Does any of that help?

bruno-b-martins:
> Were you able to see if the sinatra server received anything here?

Yes, and nothing reached it.
I've bootstrapped it in a separate shell and ran the rspec with the `include_context ...` commented out. It didn't show any requests in the logs. Check the printscreen.
Captura de ecrã 2023-09-12 231359


> As an in-between option, it looks like Resolv will also raise a timeout error if you set the timeout to 0 (or less), so we could check for that and it would get us relatively close to the integration level if this turns out to be too tricky and/or brittle to be worth it.

Unfortunately, it fails a bit sooner when we assign 0 to [timeouts](https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L951).


I believe in order to properly mock the DNS server, I'd need a deeper understanding of the `Resolv` library...

bruno-b-martins:
I've managed to get a `Resolv::TimeoutError` in the `irb` console by typing command by command through what `Resolv` does...
Captura de ecrã 2023-09-13 001044

But can't yet reproduce it in the spec file...
It might be that I'm not able to reproduce it at all due to those `rescue`s you saw.

I can keep on trying tomorrow, but it's a rough endeavor to be honest 😅

bruno-b-martins:
Ok. So, I've managed to trigger the timeout via bit of a tricky way.

Screenshot 2023-09-13 at 10 55 00

You can see that it took 2 seconds to run the spec. That's because the `dns_timeout` is set to `1` and I've set the DNS resolver with 2 `nameserver`s (1 second to try to connect to each).

bruno-b-martins:
Now, the issue with the previous test setup was that the code wasn't reaching the sinatra server.
It failed in this [recv_reply](https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L689) with a `Errno::ECONNREFUSED` and the comment below it says "No name server running? Don't wait anymore.".
This adds up because the shell that was running the sinatra server didn't get any log line when I ran the spec.

Now, why does it work with 2 nameservers set up?
When we set up only 1 nameserver config, `Resolv` creates a `Requester::ConnectedUDP` which only has 1 socket in `@socks` and sets up correctly the `select_result` variable. Then, it begins to try to receive the reply from the nameserver.
However, when we set up 2 nameserver configs, `Resolv` creates a `Requester::UnconnectedUDP` which has 2 sockets in `@socks` and doesn't set up correctly the `select_result` variable because there is no nameserver running. Thus, it retries to set that variable until the timeout is reached.

You can check this ☝️ logic in https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L678-L695.

bruno-b-martins:
Although it's not waiting for the DNS name server to reply back with an IP address, but rather waiting to connect successfully with the DNS name server, I'm still happy with this spec.

What do you say @geemus ?

geemus:
Thanks for the further details and explanation. Some of that resolv code definitely seems strange, though I imagine there are reasons for it to do those things. I think I agree that this seems good enough for our purposes. Thanks!

Do you need this in a release soon?

geemus (approved):
Thanks!

bruno-b-martins:
Yes, I'd appreciate if you could take the time to release it because I intend to update another gem before I make use of this in my application's code.

I can also help with releasing it if you point me in the right direction.

geemus:
Thanks, I had another thing to push out as well, so I just finished v0.103.0 which should include this.

Just let me know if you need anything else and thanks again for working through this with me.

bruno-b-martins:
Thank you for all the help and being so quick to provide meaningful feedback! 🙇

Issue #832, opened 2023-09-15
DefaultResolver gets ignored with excon 0.103.0


robinmoneybird:
[From what I can tell](https://github.com/excon/excon/commit/9d167537ad927bb0121c25c3511c090434082005#diff-a7562e6a39620d09af94e896c18d6c7b97539790fc63e696f68e83b480a5dfe4L124-R127), is that after the changes made in #830, which are included in release `0.103.0`, the Resolv `DefaultResolver` is ignored.

This change makes it impossible to use an overwritten default Resolv config (for example, in my use-case with a Rails initializer):

```ruby
Resolv::DefaultResolver.replace_resolvers(
  [
    Resolv::Hosts.new,
    Resolv::DNS.new,
    Resolv::DNS.new(nameserver: ['127.0.0.1'])
  ]
)
```

Would it be possible to use the `Resolv::DefaultResolver` and add the DNS timeouts config to the predefined `Resolv::DNS` instances, or something that would make using a custom Resolv configuration possible?

geemus:
Thanks for the detailed report, that is definitely an unintended consequence of our recent change.

I guess my inclination might be to at least partially revert the recent timeout related interface changes, and document for folks that they can instead use `replace_resolvers` as you have (with a clear example of setting the timeout on the DNS portions). I think that would work for everyone, but unfortunately is probably a breaking change from what we've just done (we could potentially do some deprecation-warning workaround before removing it I suspect). It's a bit more cumbersome interface-wise than the simple option and the backwards incompatibility is unfortunate, but it also seems much more flexible and futureproof.

What do you think?

bruno-b-martins:
Hey there,

I agree with @robinmoneybird that a revert outweighs the incompatibility issue introduced in `v.0.103.0`.

Since this change was introduced less than 1 week ago, it might be that nearly no one is using it yet.
I added it to [avro_turf](https://github.com/dasch/avro_turf/blob/master/lib/avro_turf/confluent_schema_registry.rb#L39) last week as well, but probably most people simply ignored it by now.


Thinking on @robinmoneybird's suggestion of setting the resolvers outside of `excon`, I get mixed feelings.
I see how it would work, but we would be transpiring the configuration of a dependency of `excon` to whatever application uses `excon`.
And, in my use case, I would have an initializer in my application to set up a gem (Resolv) that is used by a gem (Excon) that is used by a gem (AvroTurf). That configuration would also apply to the whole application, not only to Excon.
[...]

PR #834, opened 2023-09-28
Add resolv_resolver optional parameter

bruno-b-martins:
## Goal

As a result of #832 , we came up with the solution of passing the Resolv resolvers directly into Excon.

## Description

This PR adds a new `resolv_resolver` input parameter with a default `nil` value and uses it in `Excon::Socket`.
If the `dns_timeouts` input parameter is sent, it overrides the `resolv_resolver` parameter.
This is an in-between state, in order to deprecate `dns_timeouts`.


bruno-b-martins:
@geemus , I think this is ready for review.

I've went down the path of deprecating `dns_timeouts` but if this parameter is something you wish `Excon` to keep supporting (in order to make it easier to use timeouts) I can adapt the code somehow.

lib/excon/socket.rb
```diff
@@ -121,9 +121,15 @@ def connect
         family = @data[:proxy][:family]
       end
 
-      dns_resolver = Resolv::DNS.new
-      dns_resolver.timeouts = @data[:dns_timeouts]
-      resolver = Resolv.new([Resolv::Hosts.new, dns_resolver])
+      resolver = @data[:resolv_resolver] || Resolv.new
+
+      # Deprecated
+      if @data[:dns_timeouts]
```
geemus, line 127: Could you add a printed warning about the deprecation as well, which suggests using resolv_resolvers instead? ie something like https://github.com/excon/excon/blob/85556aeb4af10e94f876a8cbdb764f0377fa0d3a/lib/excon.rb#L76

geemus (commented):
I think deprecation makes sense and this approach looks reasonable to me. I'd like to add a printed warning for the deprecation, similar to what we've done elsewhere, but then I think it should be good to go. Thanks!

geemus (approved):
LGTM, thanks!

Jack12816:
Hey there! 

What happened to this idea from @geemus at https://github.com/excon/excon/issues/832#issuecomment-1725784788? 

> By default we could configure excon then to still use `Resolv::DefaultResolver` or instead use the specified resolvers. 

Now it's always set to `Resolv.new` instead of `Resolv::DefaultResolver` via:

```ruby
resolver = @data[:resolv_resolver] || Resolv.new
```

I'm asking because I'm also affected by this change documented in https://github.com/excon/excon/issues/832#issuecomment-1725784788. :)

bruno-b-martins:
Hey 👋 

It's been a while and I don't exactly recall why I put `Resolv.new` instead of `Resolv::DefaultResolver`.

Anyway, since `Resolv.new` is the same as `self.new` in https://github.com/ruby/resolv/blob/master/lib/resolv.rb#L3371, the result of using `Resolv::DefaultResolver` here should be the same, right? Or am I missing something?

Jack12816:
Wanna cry? It isn't :'(

```ruby
require 'resolv'

Resolv::DefaultResolver.replace_resolvers(
  [
    Resolv::Hosts.new,
    Resolv::MDNS.new,
    Resolv::DNS.new
  ]
)

pp Resolv::DefaultResolver
# #<Resolv:0x000055f2d6089a70
#  @resolvers=
#   [#<Resolv::Hosts:0x000055f2d6089688
#     @filename="/etc/hosts",
#     @initialized=nil,
#     @mutex=#<Thread::Mutex:0x000055f2d6089660>>,
#    #<Resolv::MDNS:0x000055f2d6089638
#     @config=
#      #<Resolv::DNS::Config:0x000055f2d60895c0
#       @config_info=
#        {:nameserver_port=>[["224.0.0.251", 5353], ["ff02::fb", 5353]]},
#       @initialized=nil,
#       @mutex=#<Thread::Mutex:0x000055f2d6089598>,
#       @timeouts=nil>,
#     @initialized=nil,
#     @mutex=#<Thread::Mutex:0x000055f2d60895e8>>,
#    #<Resolv::DNS:0x000055f2d6089570
#     @config=
#      #<Resolv::DNS::Config:0x000055f2d6089520
#       @config_info=nil,
#       @initialized=nil,
#       @mutex=#<Thread::Mutex:0x000055f2d60894f8>,
#       @timeouts=nil>,
#     @initialized=nil,
#     @mutex=#<Thread::Mutex:0x000055f2d6089548>>]>

pp Resolv.new
# #<Resolv:0x000055f2d60d3f30
#  @resolvers=
#   [#<Resolv::Hosts:0x000055f2d60d3f08
#     @filename="/etc/hosts",
#     @initialized=nil,
#     @mutex=#<Thread::Mutex:0x000055f2d60d3ee0>>,
#    #<Resolv::DNS:0x000055f2d60d3eb8
#     @config=
#      #<Resolv::DNS::Config:0x000055f2d60d3e68
#       @config_info=nil,
#       @initialized=nil,
#       @mutex=#<Thread::Mutex:0x000055f2d60d3e40>,
#       @timeouts=nil>,
#     @initialized=nil,
#     @mutex=#<Thread::Mutex:0x000055f2d60d3e90>>]>
```

geemus:
@Jack12816 Thanks, we probably just lost that suggestion in the shuffle of trying to get everything sorted out and didn't realize (or forgot) about the issue you pointed out. I've been preoccupied with family and holiday stuff, but hope to get back to review your new stuff and release if needed early to mid next week.

@bruno-b-martins Thanks again for your help here, we made good progress and I think this looks like another refinement on top.

geemus:
Apologies for my delays, I'm finally getting back to this. Just want to be really explicit about things for posterity.

As @bruno-b-martins suggests, `Resolve::DefaultResolver` does get set as `Resolv.new` at require time. In many cases this will work just fine and whether we use one or the other would be interchangeable. However, as @Jack12816 points out, if you then use use `Resolv::DefaultResolver.replace_resolvers` it will change the value of `Resolv::DefaultResolver` but not the value of `Resolv.new`. In that case, excon doesn't respect the change from `replace_resolvers` without a change to our code.

#846 changes the code to default to `Resolve::DefaultResolver` instead of `Resolv.new`. This should still work fine in the base case, but also ensures that we respect changes that might come from `replace_resolvers` as well.

Hope that clarifies, had to take a bit to reread and irb things to confirm. Thanks to you both for your work here.

bruno-b-martins:
Thanks for explaining it further @geemus . I too wasn't able to focus on this until now and I'm glad everything worked out 😅 

Great work @Jack12816 💪

Issue #596, opened 2016-08-31
Overall request level timeout


colin-jack:
There does not seem to be a way to apply an overall timeout to a request.

I have looked at connect_timeout, read_timeout, and write_timeout but what I really want is an overall timeout. Timeout::timeout seemed one option but appears to have is own issues which I would prefer to avoid

geemus:
`Timeout::timeout` does in fact have it's own issues. Unfortunately we don't currently have an overall timeout setup, as the book-keeping gets a little hairy. In most cases we have found ways to make the existing timeouts work well enough. Could you detail your situation a bit more so that I can understand where you are having trouble? Thanks!

mahemoff:
I believe `read_timeout`refers to idle time rather than total time. For example, pointing to a streaming URL will never terminate:

```
Excon.new('http://bbcmedia.ic.llnwd.net/stream/bbcmedia_radio1_mf_p', connect_time: 5, read_timeout: 5).get
```

The read timeout seems to say "raise an exception if we don't receive anything for 5 seconds" rather than "raise an exception if we are still reading after 5 seconds".

I'd like a way to guarantee this will take no more than 10 seconds. In my case, I would just want an option like `total_read_timeout`.

colin-jack:
Hi,

Yup that's what we need too. Ideally though we would actually have a total_request_timeout which covers connecting, writing any request body, then reading response.

Thanks,

Colin

colin-jack:
@geemus

Thanks for replying. 

On situation, we are calling what is essentially a third party http API and to provide decent user experience we would like to provide timeouts in each request. These timeouts are from a user's perspective so really we would like them to cover the entire request.

geemus:
Understood. It seems like a reasonable thing to have, but I'm afraid that I'm unlikely to have time to work toward it in the near term. I worry a little bit about how messy it will be to support also, but regardless I would be happy to support the effort if somebody would like to give it a try. Thanks!

mahemoff:
Not sure if I'll have time in the short term either, but looks like the main logic [is here](https://github.com/excon/excon/blob/789218e4786f90ebc3201b89d7b00000aee9e411/lib/excon/socket.rb#L256):

```
  def select_with_timeout(socket, type)
      select = case type
      when :connect_read
        IO.select([socket], nil, nil, @data[:connect_timeout])
      when :connect_write
        IO.select(nil, [socket], nil, @data[:connect_timeout])
      when :read
[...]

PR #835, opened 2023-11-14
Add a request timeout to Excon for non-blocking sockets

misalcedo:
Closes https://github.com/excon/excon/issues/596

## Approach
Today, blocking connections do not have connect or write timeouts (only read timeouts). Therefore, the request timeout only applies to non-blocking connections. The expectation seems to be that for a blocking connection, the caller is wrapping `Excon` requests in a `Timeout.timeout` block.

One of the goals of this implementation is to minimize the cost to requests not using a timeout. Therefore, the clock is not queried and the deadline is not checked unless a timeout is defined.

I used `Process.clock_gettime(Process::CLOCK_MONOTONIC)` instead of `Time.now` to avoid the added cost of allocating a Time object and to ensure that we track request deadlines correctly (i.e. not affected by clock sync).

For the timeout kinds, I chose to use a Hash mapping `type` to timeout instead of a method because a micro-benchmark I wrote showed the hash approach was ~2x faster.

## Testing
All existing tests pass and I added some tests for both blocking and non-blocking sockets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,9 @@ connection.request(:read_timeout => 360)
 # set longer write_timeout (default is 60 seconds)
 connection.request(:write_timeout => 360)
 
+# set a request timeout in seconds (default is no timeout)
+connection.request(:timeout => 5)
+
 # Enable the socket option TCP_NODELAY on the underlying socket.
 #
 # This can improve response time when sending frequent short
--- a/README.md
+++ b/README.md
@@ -179,8 +179,9 @@ connection.request(:read_timeout => 360)
 # set longer write_timeout (default is 60 seconds)
 connection.request(:write_timeout => 360)
 
-# set a request timeout in seconds (default is no timeout)
-connection.request(:timeout => 5)
+# set a request timeout in seconds (default is no timeout).
+# the timeout may be an integer or a float to support sub-second granularity (e.g., 1, 60, 0.005 and 3.5).
+connection.request(:timeout => 0.1) # timeout if the entire request takes longer than 100 milliseconds
 
 # Enable the socket option TCP_NODELAY on the underlying socket.
 #
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -315,7 +315,7 @@ module Excon
       timeout = @data[OPERATION_TO_TIMEOUT[type]]
 
       if @data.include?(:deadline)
-        remaining = check_deadline!
+        remaining = request_time_remaining
 
         if remaining < timeout
           timeout_kind = :request
@@ -345,9 +345,9 @@ module Excon
       end
     end
 
-    # Checks whether we have exceeded the request deadline.
-    # Returns the remaining time in seconds until we reach the deadline.
-    def check_deadline!
+    # Returns the remaining time in seconds until we reach the deadline for the request timeout.
+    # Raises an exception if we have exceeded the request timeout's deadline.
+    def request_time_remaining
       now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
       deadline = @data[:deadline]
       
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -342,7 +342,7 @@ module Excon
       now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
       deadline = @data[:deadline]
       
-      raise raise(Excon::Errors::Timeout.new("request timeout reached")) if now >= deadline
+      raise raise(Excon::Errors::Timeout.new('request timeout reached')) if now >= deadline
 
       remaining  = deadline - now
 
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -9,8 +9,6 @@ module Excon
 
     attr_accessor :data
 
-    REQUEST_TIMEOUT_KIND = "request"
-
     # read/write drawn from https://github.com/ruby-amqp/bunny/commit/75d9dd79551b31a5dd3d1254c537bad471f108cf
     CONNECT_RETRY_EXCEPTION_CLASSES = if defined?(IO::EINPROGRESSWaitWritable) # Ruby >= 2.1
       [Errno::EINPROGRESS, IO::EINPROGRESSWaitWritable]
@@ -346,7 +344,7 @@ module Excon
 
       remaining  = deadline - now
 
-      remaining < timeout ? [REQUEST_TIMEOUT_KIND, remaining] : [type, timeout]
+      remaining < timeout ? [:request, remaining] : [type, timeout]
     end
   end
 end
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -9,6 +9,8 @@ module Excon
 
     attr_accessor :data
 
+    REQUEST_TIMEOUT_KIND = "request"
+
     # read/write drawn from https://github.com/ruby-amqp/bunny/commit/75d9dd79551b31a5dd3d1254c537bad471f108cf
     CONNECT_RETRY_EXCEPTION_CLASSES = if defined?(IO::EINPROGRESSWaitWritable) # Ruby >= 2.1
       [Errno::EINPROGRESS, IO::EINPROGRESSWaitWritable]
@@ -245,7 +247,7 @@ module Excon
     end
 
     def read_block(max_length)
-      @socket.read(max_length)
+              @socket.read(max_length)
     rescue OpenSSL::SSL::SSLError => error
       if error.message == 'read would block'
         select_with_timeout(@socket, :read) && retry
@@ -294,7 +296,7 @@ module Excon
     end
 
     def write_block(data)
-      @socket.write(data)
+              @socket.write(data)
     rescue OpenSSL::SSL::SSLError, *WRITE_RETRY_EXCEPTION_CLASSES => error
       if error.is_a?(OpenSSL::SSL::SSLError) && error.message != 'write would block'
         raise error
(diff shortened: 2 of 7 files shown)


lib/excon/socket.rb
```diff
@@ -304,17 +311,30 @@ def write_block(data)
...
+        if remaining < timeout
+          timeout_kind = :request
+          timeout = remaining
+        end
+      end
+
       select = case type
       when :connect_read
-        IO.select([socket], nil, nil, @data[:connect_timeout])
+        IO.select([socket], nil, nil, timeout)
```
geemus, line 328: Maybe we could simplify this, and the lookup above. An idea I had would be to have a local variable here, something like `request_timeout` which would either be nil or a float if there is time remaining (much as you have above). The difference would be that here in the `IO.select` we could have something like `request_timeout || @data[:connect_timeout]`. That feels like it should be mostly the same, but maybe a little easier to read and shorter. What do you think?

tests/rackups/timeout.ru
```diff
@@ -4,6 +4,10 @@ class App < Sinatra::Base
   set :environment, :production
   enable :dump_errors
 
+  get('/') do
```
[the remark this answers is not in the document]
geemus: Conversely, there is some nice simplicity in just using one rackup throughout, but in that case I still think it might be nice to make it be something that is more explicit and hopefully obvious. So maybe we could have this be something like `get('no-timeout')` or something? That way it hopefully it easier to remember and see what is expected. What do you think?

geemus (changes requested):
Looking great overall, I just added a few small comments that I think could polish it up even more. Just let me know if you have questions and I'm certainly happy to discuss all of it. Thanks again for taking this on!

README.md
```diff
@@ -179,6 +179,9 @@ connection.request(:read_timeout => 360)
 # set longer write_timeout (default is 60 seconds)
 connection.request(:write_timeout => 360)
 
+# set a request timeout in seconds (default is no timeout)
```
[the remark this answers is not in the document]
misalcedo: Good point, I updated the readme with some more details and updated the example to be a 100 millisecond timeout.

tests/rackups/timeout.ru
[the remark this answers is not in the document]
misalcedo: I didn't use basic because I wasn't sure I wanted to run 2 servers for these tests. I like the `no-timeout` suggestion.

lib/excon/socket.rb
misalcedo: I used the `timeout` variable in order to keep some sort of symmetry between the timeout and the kind of timeout. Since we need to update the kind used in the exception messsage if we use the request timeout.

At first I had a conditional in each branch to choose the timeout and kind to use, but that seemed too repetitive. I also tried returning the kind and timeout from the check deadline method but I didn't like the multi-return pattern. Something like `timeout, kind = check_deadline!(@data[:connect_timeout], :connect_read)` in each branch.

Part of the complexity is that we only want to use the request timeout if it is smaller than the operation's timeout. We don't want to use the request timeout just because one is defined, which is what we would get with `request_timeout || @data[:connect_timeout]`.

misalcedo: This is what I think the select method would look like with your suggestion (please let me know if I misunderstood):

```ruby
    def select_with_timeout(socket, type)
      timeout_kind = type
      request_timeout = check_deadline! if @data.include?(:deadline)

      select = case type
      when :connect_read
        if request_timeout && request_timeout < @data[:connect_timeout]
          timeout_kind = type
        else
          request_timeout = nil
        end

        IO.select([socket], nil, nil, request_timeout || @data[:connect_timeout])
      when :connect_write
        # ...
        IO.select(nil, [socket], nil, @data[:connect_timeout])
      when :read
        # ...
        IO.select([socket], nil, nil, @data[:read_timeout])
      when :write
        # ...
        IO.select(nil, [socket], nil, @data[:write_timeout])
      end
      select || raise(Excon::Errors::Timeout.new("#{type} timeout reached"))
    end
```

misalcedo: If it wasn't for the kind things would be a little bit simpler:

```ruby
    def select_with_timeout(socket, type)
      request_timeout = check_deadline! if @data.include?(:deadline)

      select = case type
      when :connect_read
        if !request_timeout.nil? && request_timeout >= @data[:connect_timeout]
          request_timeout = nil
        end

        IO.select([socket], nil, nil, request_timeout || @data[:connect_timeout])
      when :connect_write
        # ...
        IO.select(nil, [socket], nil, @data[:connect_timeout])
      when :read
        # ...
        IO.select([socket], nil, nil, @data[:read_timeout])
      when :write
        # ...
        IO.select(nil, [socket], nil, @data[:write_timeout])
      end
      select || raise(Excon::Errors::Timeout.new("#{type} timeout reached"))
    end
```

misalcedo: At one point this is the solution I had. I didn't like it because we always paid a penalty of allocating those arrays and the method call regardless of whether a timeout was defined or not.

```ruby
    def select_with_timeout(socket, type)
      timeout_kind = type

      select = case type
      when :connect_read
        timeout_kind, timeout = check_deadline!(type, @data[:connect_timeout])
        IO.select([socket], nil, nil, timeout)
      when :connect_write
        timeout_kind, timeout = check_deadline!(type, @data[:connect_timeout])
        IO.select(nil, [socket], nil, timeout)
      when :read
        timeout_kind, timeout = check_deadline!(type, @data[:read_timeout])
        IO.select([socket], nil, nil, timeout)
      when :write
        timeout_kind, timeout = check_deadline!(type, @data[:write_timeout])
        IO.select(nil, [socket], nil, timeout)
      end

      select || raise(Excon::Errors::Timeout.new("#{timeout_kind} timeout reached"))
    end

    # Checks whether we have exceeded the request deadline.
    # Returns the remaining time in seconds as Numeric.
    def check_deadline!(type, timeout)
      return [type, timeout] unless @data.include?(:deadline)

      now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
      deadline = @data[:deadline]
      
      raise raise(Excon::Errors::Timeout.new('request timeout reached')) if now >= deadline

      remaining  = deadline - now

      remaining < timeout ? [:request, remaining] : [type, timeout]
    end
```

geemus: Those don't exactly line up with what I had in mind, but I think what I was thinking wouldn't work anyway now that I understand things a bit better from your explanation. That being said, I think it might be good to add some of that explanation as a comment to hopefully make it clearer. I'll add another comment in a second with what I mean.

lib/excon/socket.rb
```diff
@@ -324,5 +344,16 @@ def unpacked_sockaddr
         raise
       end
     end
+
+    # Checks whether we have exceeded the request deadline.
+    # Returns the remaining time in seconds until we reach the deadline.
+    def check_deadline!
```
geemus, line 350: It might be clearer to name this based on our intention, rather than what it does. So rather than saying it checks the deadline, maybe something like `request_time_remaining` would make it more obvious?

lib/excon/socket.rb
```diff
@@ -304,17 +311,30 @@ def write_block(data)
     end
 
     def select_with_timeout(socket, type)
+      timeout_kind = type
+      timeout = @data[OPERATION_TO_TIMEOUT[type]]
+
+      if @data.include?(:deadline)
```
geemus, line 317: I think it would be good to add a bit of a comment here to explain the logic/intention of this chunk of code, since it's important (and since I don't think I really fully understood it during the earlier review), and maybe the slight rename suggested below. I was thinking something like: `If a request timeout is set, use the smaller of the time remaining before the request timeout and the timeout for the type of select`.

geemus (changes requested):
Hey, thanks for working through those changes and explaining why the other parts were that way, it definitely helped me better understand and see why my idea wouldn't really work.

That said, I had a couple further small tweaks that I hope will make more of your explanation obvious within the code itself. Would love to know what you think, and I'm certainly happy to further discuss. Otherwise I think it's looking good, thanks again!

lib/excon/socket.rb
misalcedo: That name makes sense to me. I renamed to `request_time_remaining` and updated the doc comment to make it clearer that we are checking the time remaining until we hit the request's timeout.

lib/excon/socket.rb
misalcedo: Thanks for pointing out that the code was not obvious. I added 2 comments: one above the deadline check to explain why we check for a deadline, the other above the check to see if we should use the request timeout (basically my own wording on your suggestion above).

I also renamed the `remaining` local variable to `request_timeout` to make it clearer what the value represents.

misalcedo:
Thank you for the thorough reviews. I really appreciate you pointing out where the code is less obvious than I expected it to be. I think your suggestions make sense and I went ahead and implemented both.

lib/excon/socket.rb
geemus: Looks good, thanks!

geemus:
Of course, thanks for your efforts and for working through it with me to make it the best we could.

misalcedo:
Looking at the failed test, I think this is a race condition because the timeouts are both so close to each other. I updated the values so there would be a much bigger gap between the timeouts.

geemus:
Thanks, I was just looking at that as well. Hopefully that settles it.

geemus (approved):

geemus:
Thanks!

I'm a bit busy this week with some holiday related things, so I plan to defer a release until early next week when I'll be more available should anything go wrong. Thanks again for your help and for working through this with me to refine it.

geemus:
Released in v0.105.0.

PR #838, opened 2023-11-22
Performance improvement on nonblocking sockets

misalcedo:
The goal of this change is to improve the performance of excon while maintaining a low number of allocations.

This change is similar in nature to what `Net::HTTP` does in https://github.com/ruby/ruby/blob/v3_2_2/lib/net/protocol.rb#L166 . 

The following is a zipped archive of 3 ruby-prof profiles measuring allocations. From the profiles, it is evident the number of allocations in OpenSSL calls and Socket calls remains very similar (still significantly less than v0.71.0).

[Excon.zip](https://github.com/excon/excon/files/13433081/Excon.zip)

Results from my benchmark show this is a marked improvement from the latest version:
```
Modified:      377.9 i/s
v0.104.0:      258.3 i/s
v0.71.0:       364.4 i/s
```

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -213,11 +213,19 @@ module Excon
       begin
         if max_length
           until @backend_eof || @read_buffer.length >= max_length
-            @read_buffer << @socket.read_nonblock(max_length - @read_buffer.length)
+            if @read_buffer.empty?
+              @read_buffer = @socket.read_nonblock(max_length, @read_buffer)
+            else
+              @read_buffer << @socket.read_nonblock(max_length - @read_buffer.length)
+            end
           end
         else
-          while !@backend_eof
-            @read_buffer << @socket.read_nonblock(@data[:chunk_size])
+          until @backend_eof
+            if @read_buffer.empty?
+              @read_buffer = @socket.read_nonblock(@data[:chunk_size], @read_buffer)
+            else
+              @read_buffer << @socket.read_nonblock(@data[:chunk_size])
+            end
           end
         end
       rescue OpenSSL::SSL::SSLError => error
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -53,14 +53,16 @@ module Excon
       @nonblock = data[:nonblock]
       @port ||= @data[:port] || 80
       @read_buffer = String.new
+      @read_offset = 0
       @eof = false
       @backend_eof = false
+
       connect
     end
 
     def read(max_length = nil)
       if @eof
-        return max_length ? nil : ''
+        max_length ? nil : ''
       elsif @nonblock
         read_nonblock(max_length)
       else
@@ -71,20 +73,22 @@ module Excon
     def readline
       if @nonblock
         result = String.new
-        block = @read_buffer
-        @read_buffer = String.new
+        block = consume_read_buffer
 
         loop do
           idx = block.index("\n")
+
           if idx.nil?
             result << block
           else
-            result << block.slice!(0, idx+1)
-            add_to_read_buffer(block)
+            result << block[..idx]
+            add_to_read_buffer(block, idx)
             break
           end
+
           block = read_nonblock(@data[:chunk_size]) || raise(EOFError)
         end
+
         result
       else # nonblock/legacy
         begin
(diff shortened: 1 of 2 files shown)


geemus:
@misalcedo Thanks, I do hope to review this soon, but have been a bit behind with holiday travel and catching up on other things.  So thanks for your patience also in the interim.

misalcedo:
No worries, take as much time as you need.

lgo:
@misalcedo, this is absolutely fantastic - thank you! I'll also try to take a read through and see what I can do to test this internally at Stripe, but I'm not sure I'll have time until closer to the holidays. Embarrassingly, we still have not bumped versions after having internal breakages from #796 (🤞 resolved by v0.95.0).

I recently made some observations that the current code, while much lower for allocs, did negatively impact total allocated memory. The header string would be re-allocated as each line was processed, getting smaller and smaller. (My naive understanding is that the former is much worse than the latter)

From a quick skim of your changes, I believe you may have also addressed that but I'll do a more thorough read later.

misalcedo:
The more eyes on it the better. I haven't tried to do any memory profiling, but that should be fairly simple since I setup the benchmark with Ruby-prof. As long as you are not using Ruby 3 that is. Ruby-prof segfaults for me whenever I try to do memory profiling on Ruby 3.2.2. I think this is because the garbage collector is different, but not entirely sure.

lgo:
Yep, I've faced similar segfault issues. When I last measured this, I used https://github.com/Shopify/memory_profiler. You can  wrap a block with the profiler and get pretty granular information on (1) alloc count, (2) source of allocs, (3) size and specific strings allocated.

When I get to testing this internally, I'll also spin up the memory profiler and check. It's been almost a year since I last walked through profiling Excon but I have a bunch of notes + hacked up PRs from the last time I did it I'll look back on.

lib/excon/socket.rb
```diff
@@ -237,18 +269,33 @@ def read_nonblock(max_length)
...
+        @read_offset = @read_buffer.length
+        @eof = @backend_eof
+
+        @read_buffer[start..]
       end
-      @eof = @backend_eof && @read_buffer.empty?
-      ret
+    end
+
+    def buffer_length
```
geemus, line 297: For clarity, maybe we should think about naming this something like `buffer_remaining` (presuming I'm understanding it's intention). Right now it otherwise gets a little confusing what the difference between this and `@read_buffer.length` is. Does that make sense?

geemus: Or perhaps `buffer_available`. That being said, I guess maybe flipping the two words in either case might read better too. ie `remaining_buffer` and `available_buffer` both seem like they might be easier to read. I guess all told, I'd maybe lean toward `remaining_buffer` from those options, but I'm open to other possibilities and discussion.

lib/excon/socket.rb
```diff
@@ -237,18 +269,33 @@ def read_nonblock(max_length)
...
+          @eof = @backend_eof
+          nil
         else
-          @read_buffer.slice!(0, max_length)
+          start = @read_offset
+          readable_bytes = buffer_length
+
+          # Ensure that we can seek backwards when reading until a terminator string.
+          # The read offset must never point past the end of the read buffer.
+          @read_offset += max_length > readable_bytes ? readable_bytes : max_length
```
geemus, line 283: Relating to the comment below, if `buffer_length` was referred to as something like `remaining_buffer`, I think we could probably just refer to it directly here and things would be relatively clear (as opposed to assigning it to `readable_bytes` in order to provide clarity, or at least I'm not immediately seeing another reason for that particular allocation).

geemus (changes requested):
Apologies for the delays, I wanted to make sure I could find a time when I was fresh and focused to give this a thorough look. Overall it seems great, thanks again for your work and help. I had a couple minor suggestions about naming/cleanup that I'd like to discuss, but other than that I think it's very nearly there. Thanks!

lib/excon/socket.rb
misalcedo: That makes sense to me.

lib/excon/socket.rb
geemus: Good call. I think of those I personally like `readable_bytes` best. It says something about our intention/expectation about what we want to do with the bytes, vs `buffered_bytes`, which is accurate, but much less specific I think.

misalcedo: I went ahead and changed the name to `readable_bytes`. Let me know what you think.

lib/excon/socket.rb
misalcedo: However, I don't think a single method call is going to be the differentiator here. I changed the code to use the method instead of a variable on the stack.

geemus: yeah, probably would really be fine either way.

geemus (approved):
Thanks, I think the `readable_bytes` change does help a lot in terms of quickly and easily understanding.

geemus:
@misalcedo thanks again for the work you did here and your collaboration in polishing it.

@lgo I'll be curious to hear what you find, and would welcome any notes/code you have that might help us profile in the future (it's not something I have much experience with either). Also, just let me know if there is anything you need so that you all can get bumped to a newer version.

lgo:
> I'll be curious to hear what you find, and would welcome any notes/code you have that might help us profile in the future (it's not something I have much experience with either).

Absolutely, I'd be more than happy to share notes on the whole process and see what I can do to provide a simple or reproducible suite (maybe just tweaking @misalcedo's). 

@misalcedo - thanks again for this awesome changes as well as the benchmarks to keep this a standard!

misalcedo:
Contributing was actually pretty fun for me, so glad I could help. I played around with the memory profiler but struggled to process the vast amount of information it provided. Now I have something new to learn this weekend.

PR #844, opened 2023-12-15
fix old ruby array syntax error from #838

### LINKED ISSUE #843: syntax error, unexpected ']' in excon-0.107.0/lib/excon/socket.rb:213:
Hey, sorry to bother you, you recently released your 1.107.0 tag and this has caused some of our infrastructure that uses chef to throw the following stacktrace. (I am not a ruby dev sorry) Is this expected behaviour due to a BC and ruby version (I am guessing so) or something that could be fixed?

This caught us off guard but we will be pinning to 1.106 for the areas affected. (This is the danger not pinning to begin with!)

```

Running handlers:
[2023-12-15T15:36:00+00:00] ERROR: Running exception handlers
Running handlers complete
[2023-12-15T15:36:00+00:00] ERROR: Exception handlers complete
Chef Client failed. 0 resources updated in 12 seconds
[2023-12-15T15:36:00+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2023-12-15T15:36:00+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2023-12-15T15:36:00+00:00] FATAL: SyntaxError: /opt/chef/embedded/lib/ruby/gems/2.5.0/gems/excon-0.107.0/lib/excon/socket.rb:84: syntax error, unexpected .., expecting ']'
            result << block[..idx]
                            ^~
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/excon-0.107.0/lib/excon/socket.rb:213: syntax error, unexpected ']'
... = @read_buffer[@read_offset..]
...                              ^
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/excon-0.107.0/lib/excon/socket.rb:292: syntax error, unexpected ']'
        @read_buffer[start..]
                            ^
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/excon-0.107.0/lib/excon/socket.rb:408: syntax error, unexpected end-of-input, expecting keyword_end
```

geemus:
closes #843

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -81,7 +81,7 @@ module Excon
           if idx.nil?
             result << block
           else
-            result << block[..idx]
+            result << block[0..idx]
             rewind_read_buffer(block, idx)
             break
           end
@@ -210,7 +210,7 @@ module Excon
 
     # Consume any bytes remaining in the read buffer before making a system call.
     def consume_read_buffer
-      block = @read_buffer[@read_offset..]
+      block = @read_buffer[@read_offset..-1]
 
       @read_offset = @read_buffer.length
 
@@ -289,7 +289,7 @@ module Excon
         @read_offset = @read_buffer.length
         @eof = @backend_eof
 
-        @read_buffer[start..]
+        @read_buffer[start..-1]
       end
     end
 


geemus:
NOTE: I originally pushed a commit adding Ruby 2.5 to the test matrix so I could red/green test this (which I successfully did). I would prefer to not add unsupported Ruby versions back into the CI rotation at present though, so to do that and keep history cleaner I removed that commit and force pushed. Just wanted to add this context for clarity/posterity.

dentarg:
Should excon specify [`required_ruby_version`](https://guides.rubygems.org/specification-reference/#required_ruby_version) in the gemspec? Then users wouldn't get a version of excon their Ruby can't handle.

I suspect continuing without `required_ruby_version` in the gemspec and not having old Ruby versions in CI will just make these types of errors pop up again and again.

geemus:
@dentarg great idea. I didn't know about that option (or at least forgot about it). Definitely seems like it could head off this issue, and allow us to use newer syntax and stuff without having to worry so much about breakage. I guess then that we should just set the required version to match whatever the oldest version still under CI would be? I think that should get us what we would want. I'd welcome a quick PR if you have a moment, but otherwise I will try to work it up in the near future to give it a try. I don't feel like it probably is so urgent to warrant a release, but it would be good to have it in there for whenever the next release does come up.

dentarg:
@geemus Sure, https://github.com/excon/excon/pull/847

geemus:
@dentarg Awesome, thanks!

PR #847, opened 2024-01-04
Require at least Ruby 2.7 for future excon releases

dentarg:
To avoid problems like https://github.com/excon/excon/issues/843

Discussed at https://github.com/excon/excon/pull/844#issuecomment-1876931461


geemus (approved):
LGTM, thanks!

PR #850, opened 2024-04-21
Separate timeout errors by type

maxim:
Excon used to raise `Excon::Error::Timeout` for any type of timeout. This PR splits this error into multiple specific ones, with the following inheritance hierarchy:

```
Timeout
|_ReadTimeout
|_WriteTimeout
|_ConnectTimeout
  |_ConnectReadTimeout
  |_ConnectWriteTimeout
```

Since all timeout errors inherit from `Excon::Error::Timeout`, we maintain backwards compatibility.

A bonus benefit is that even when we raise a generic "request timeout reached" we are still raising the correct _type_ of exception, such that a user can tell the phase of the request in which timeout actually occurred. This makes the request timeout feature more powerful.

## Breaking changes

- The `DEFAULT_RETRY_ERRORS` no longer includes read and write timeouts, only connect. It seems to me the safer option, but I wonder how you feel about making this change here.
- I did change exception "connect timeout reached" to say "connect [read/write] timeout reached" instead. We probably shouldn't consider it a breaking change, but I can see it both ways.

## Questions

We could probably use more tests. If you think we should, where should I add them? (test, tests, or spec)?

### DIFF
--- a/lib/excon/error.rb
+++ b/lib/excon/error.rb
@@ -51,7 +51,7 @@ or:
     class ResponseParse < Error; end
 
     class Timeout < Error
-      def self.by_type(type, human_name)
+      def self.by_type(type, human_name = type)
         case type
         when :connect_read
           Excon::Errors::ConnectReadTimeout.described_as(human_name)


lib/excon/error.rb
```diff
@@ -48,9 +48,33 @@ def initialize(socket_error = Excon::Error.new)
...
+        when :read
+          Excon::Errors::ReadTimeout.described_as(human_name)
+        when :write
+          Excon::Errors::WriteTimeout.described_as(human_name)
+        end
+      end
+
+      def self.described_as(timeout_kind)
+        new("#{timeout_kind.to_s.tr('_', ' ')} timeout reached")
+      end
```
maxim, line 69: I found this to be the most elegant way to give us a little flexibility while maintaining DRY-ness, and keeping exceptions completely vanilla otherwise. (Not overwriting their `initialize`, etc).

lib/excon/socket.rb
```diff
@@ -364,6 +364,9 @@ def select_with_timeout(socket, type)
       if @data.include?(:deadline)
         request_timeout = request_time_remaining
 
+        # If request timeout is already reached, there's no need to proceed.
+        raise(Excon::Errors::Timeout.by_type(type, 'request')) if request_timeout <= 0
```
maxim, line 368: I wanted this exception to be of the correct type, but instead of passing the type down to `request_time_remaining` (where this exception is raised), I figured it makes more sense to extract the raise into the caller. What do you think?

lib/excon/error.rb
```diff
@@ -48,9 +48,33 @@ def initialize(socket_error = Excon::Error.new)
 
     class InvalidHeaderKey < Error; end
     class InvalidHeaderValue < Error; end
-    class Timeout < Error; end
     class ResponseParse < Error; end
 
+    class Timeout < Error
+      def self.by_type(type, human_name)
```
maxim, line 54: Not sure if `by_type` is clear enough. It seems to make sense in caller context.

lib/excon/error.rb
```diff
@@ -48,9 +48,33 @@ def initialize(socket_error = Excon::Error.new)
...
+        when :read
+          Excon::Errors::ReadTimeout.described_as(human_name)
+        when :write
+          Excon::Errors::WriteTimeout.described_as(human_name)
+        end
+      end
+
+      def self.described_as(human_name)
+        new("#{human_name.to_s.tr('_', ' ')} timeout reached")
+      end
```
maxim, line 69: I found this to be the most elegant way to give us a little flexibility while maintaining DRY-ness, and keeping exceptions completely vanilla otherwise. (Not overwriting their `initialize`, etc).

Also, I was tempted to put `.to_s.tr(…)` into the caller, but held off on it for now. Curious what you think.

lib/excon/error.rb
```diff
@@ -48,9 +48,33 @@ def initialize(socket_error = Excon::Error.new)
...
+        when :connect_read
+          Excon::Errors::ConnectReadTimeout.described_as(human_name)
+        when :connect_write
+          Excon::Errors::ConnectWriteTimeout.described_as(human_name)
+        when :read
+          Excon::Errors::ReadTimeout.described_as(human_name)
+        when :write
+          Excon::Errors::WriteTimeout.described_as(human_name)
+        end
+      end
```
maxim, line 65: Not sure if `by_type` is clear enough. It seems to make sense in caller context.

lib/excon/socket.rb
geemus: Yeah, that seems reasonable to me. I suspect it makes it a little easier to read and understand.

lib/excon/error.rb
geemus: It seems pretty good. The only further thing that comes to mind would be to tweak it toward something like `new_by_type` since it is kind of a specialized constructor that might make it a little more obvious the intention. What do you think?

lib/excon/error.rb
geemus: Yeah, I wasn't totally sure about this part. Other than request timeouts, I think they all just use their type for the human name. And especially if we made request timeouts their own thing (as I suggest in another comment), even that distinction is no longer true. At that point, it seems like it might be simpler to drop the `human_name` and `described_as` parts and move it all up into the `by_type` method as you suggest. What do you think?

geemus:
I like it so far, offered a few suggestions as well. Thanks!

# Breaking Changes e like it should be relatively safe, though it may change behaviors for some users, those behaviors were probably somewhat dangerous to begin with.
- I wouldn't really consider changing the text of the errors as breaking. I hope at least that those particulars aren't being relied upon, but it's not unlikely that someone will have done something there.

# Tests
- Probably just in `tests` would be good for further tests, as I think that is still where the brunt of the tests live. I think it would be good to at least try and trigger each of the error cases if we can. The test stuff is a bit of a mess and it would be great to consolidate/etc. Someone started converting things and then lost steam, and I haven't had bandwidth to go back to it (plus it's hard to prioritize when what we have works well enough, albeit in a bit of a messy setup).

lib/excon/socket.rb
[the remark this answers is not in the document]
maxim: The only concern is that if we do that, we lose the benefit of being able to tell which specific timeout caused the overall request timeout. It would be powerful if I could set timeout on the entire request, but retry only on connect timeouts.

That's why I opted to raise specific timeout even for requests.

Another alternative is to repeat the whole hierarchy for Request:

```
Timeout
|_ReadTimeout
|_WriteTimeout
|_ConnectTimeout
  |_ConnectReadTimeout
  |_ConnectWriteTimeout
|_RequestTimeout
  |_RequestReadTimeout
  |_RequestWriteTimeout
  |_RequestConnectTimeout
    |_RequestConnectReadTimeout
    |_RequestConnectWriteTimeout
```

this seems a little excessive however.

Are you sure you'd want to hide the specific timeout error behind RequestTimeout?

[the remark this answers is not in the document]
maxim: Forgot to mention this detail: there is already an exception called `RequestTimeout` (code 408), which would be in conflict if we did this.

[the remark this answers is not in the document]
geemus: Oops, right...

For some context, if I recall correctly the request timeout here is harder to pin down to a particular cause in the same way.

ie I expect read timeout is specifically that I started waiting at the beginning of a read and it took longer than I wanted, so I raised. Same thing for write or connect errors.

I believe the intention with the error we are talking about here is actually a timeout for the entirety of the request. So the timer starts as soon as the request process begins and it times out as soon as the limit is surpassed. That being said, it's possible for instance that the timeout actually would occur in between the other activities during excon's own processing. ie if connect took just slightly less than the request limit, the timeout would technically occur in between connect and write. In that case, it wouldn't really be correct to say that the timout was either connect or write (and hence the generic error we see here).

Does that make sense?

All that being said, I'm not sure that "request" is a very good descriptor of these expectations. Maybe it should be something like "OverallTimeout" or "TotalTimeout" or something (I'm definitely open to suggestions). It's also possible that this is more trouble than it's worth and something like what you have done is good enough. What do you think?

I'm glad to happily surprise you in terms of responsiveness, though paradoxically I was feeling bad for how long it took to get back to you...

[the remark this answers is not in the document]
maxim: > It's also possible that this is more trouble than it's worth and something like what you have done is good enough.

Could be, but you're right that we have to think this through.

> it's possible for instance that the timeout actually would occur in between the other activities

It's true, we only check monotonic clock once per operation, so we can't know when the deadline was actually crossed. Could've been during the beginning of the current phase, the end of the previous one, or in between.

But if you think about it, for "RequestTimeout" we don't actually care exactly when timeout occurred. We only care which phases of request completed successfully at the moment of raising this error. So maybe we indeed should go with something like `TotalTimeout` (better name pending), but add a piece of metadata on it, like:

```ruby
rescue Excon::Errors::TotalTimeout => e
  puts e.completed_request_phases # => [:connect_write, :connect_read, :write]
end
```

If request timeout is not set by the user, then we fallback to specific timeout errors. I don't know the code deep enough to tell if this is easy to do, but at least it avoids the need for impossible timeout precision. How does this sound?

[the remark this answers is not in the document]
geemus: Sounds good to me. Naming wise maybe we could combine the old and new ideas, so something like `OverallRequestTimeout` or `TotalRequestTimeout`? I kind of like overall between those I think. What do you think?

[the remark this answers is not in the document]
maxim: `OverallRequestTimeout` looks like a good option. We could also consider a different angle: `RequestDurationTimeout`, `RequestSpanTimeout`. What do you think about these?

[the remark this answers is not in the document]
geemus: Oh, yeah. I really like `RequestDurationTimeout`. I think that is probably my new favorite, great idea!

geemus:
@maxim thanks for the updates.

1. I think it's fine to just have a simple timeout for the request timeout case. I'm not sure how much it even gets used vs the more specific timeouts, and if you care about specifics you can you read/write limits instead of the overall one I suppose.
2. Good question. I think that we just wrap/reraise timeout related errors in particular, and that other Socket errors are allowed to raise directly if I recall (a brief look at the code seems to confirm). I think we would probably expect to catch this as it is a timeout though. In any event, feel free to file this as an issue for later. I don't think it's needed as part of the other work, but would be good to follow up on.

maxim:
@geemus It's been awhile but I haven't abandoned this. I got connect timeout test to work, but been stuck on simulating a write timeout. This is my latest attempt.

```ruby
context 'when write timeout is triggered' do
  before do
    Thread.new do
      server = TCPServer.new('127.0.0.1', 5678)

      loop do
        client = server.accept
        puts 'Client connected'
        p "Received #{client.recv(1)}"
        p 'Waiting 3 seconds...'
        sleep 3
        p "Received #{client.read}"
        client.close
      end
    end
  end

  let(:conn) do
    Excon::Connection.new(host: '127.0.0.1',
                          hostname: '127.0.0.1',
                          nonblock: false,
                          port: 5678,
                          scheme: 'http',
                          timeout: nil,
                          connect_timeout: 5,
                          write_timeout: 0.01,
                          read_timeout: 5)
  end

  it 'raises WriteTimeout error' do
    expect { conn.request(path: '/') }.to raise_error(Excon::Error::WriteTimeout, 'write timeout reached')
  end
end
```

In the `conn.request(path: '/')` the client is trying to write the following message to the server:

```
GET / HTTP/1.1\r\nHost: 127.0.0.1:5678\r\nUser-Agent: excon/0.110.0\r\nAccept: */*\r\n\r\n
```

The server (the thread on top) only accepts 1 byte (`client.recv(1)`), which is the first letter `G`, then sleeps for 3 seconds before accepting the rest of the message.

I thought this has gotta be the way to trigger a write timeout, because the client is set to `0.01` write_timeout, but won't be able to finish writing for another 3 seconds. However, instead I keep getting the read timeout every time.

```
Client connected
"Received G"
"Waiting 3 seconds..."
"Received ET / HTTP/1.1\r\nHost: 127.0.0.1:5678\r\nUser-Agent: excon/0.110.0\r\nAccept: */*\r\n\r\n"
      raises WriteTimeout error (FAILED - 1)

Failures:

  1) Excon::Connection non-blocking connection when write timeout is triggered raises WriteTimeout error
     Failure/Error: expect { conn.request(path: '/') }.to raise_error(Excon::Error::WriteTimeout, 'write timeout reached')

       expected Excon::Error::WriteTimeout with "write timeout reached", got #<Excon::Error::ReadTimeout: read timeout reached> with backtrace:
         # ./lib/excon/socket.rb:99:in `rescue in readline'
         # ./lib/excon/socket.rb:94:in `readline'
         # ./lib/excon/response.rb:73:in `block in parse'
         # ./lib/excon/response.rb:72:in `parse'
         # ./lib/excon/middlewares/response_parser.rb:7:in `response_call'
         # ./lib/excon/connection.rb:466:in `response'
         # ./lib/excon/connection.rb:297:in `request'
         # ./spec/requests/timeout_spec.rb:110:in `block (5 levels) in <top (required)>'
         # ./spec/requests/timeout_spec.rb:110:in `block (4 levels) in <top (required)>'
     # ./spec/requests/timeout_spec.rb:110:in `block (4 levels) in <top (required)>'
```

Haven't cracked this yet. If you have any ideas, would be happy to hear, but for now I keep digging in this direction.

maxim:
I imagine it's probably because the only way to trigger a write timeout is to somehow make the socket not ready to accept a write, not to delay writing once it already started.

geemus:
@maxim thanks for the updates. Yeah, I'm not sure about the expectations for write timeouts. So I'm not sure if it needs to completely block, or not. What you suggested sounds like a reasonable possibility to me, but I definitely don't know for sure. I'm also not sure how to test this one, maybe we'll have to accept that there isn't a great way to test it though and move forward anyway.

Issue #787, opened 2022-07-13
can't stub request


Asdfif:
I stub request `Excon.stub({url: 'http://example.com'}, {:body => 'body', :status => 200})` 
Then  i send request  `Excon.get('http://example.com', mock: true)`
And i got an error `Excon::Error::StubNotFound: no stubs matched`
Why? I can't find answer on stack over flow.
Help me please

geemus:
@Asdfif - Great question, I think you may have identified an edge case where the matching to stubs is overly strict that we should probably fix up. When that stub is created, it parses the URL and it looks like it interprets the path as being a blank string, whereas when the call is made it defaults to a path of `'/'`. Since blank string and `/` are not the same, it doesn't match (although in this case I would argue that it really ought to.

I can work on a patch some time in the near future, but in the mean time you can also work around this problem by being explicit with the path when you create the stub (by adding a `/`). So if you stub the request this way, it appears to work as you originally expected: `Excon.stub({url: 'http://example.com/'}, {:body => 'body', :status => 200})`.

Hope that helps, let me know if you have questions and I'll update this if/when I get a chance to fix and release.

Asdfif:
@geemus  It works. Thank you!

Asdfif:
@geemus  but if i need to stub request like this `http://example.com/:id`. how should i do it?

Asdfif:
i try to use Regexp`Excon.stub({url: /https?:\/\/example?.com\// }, {:body => 'body', :status => 200})` but it raise an error `URI::InvalidURIError: bad URI(is not URI?)`

geemus:
@Asdfif great question. The stubs are setup so that if you specify a nil value, it will match anything. So that might be the easiest. So one option would be something like:

`Excon.stub({host: 'example.com', scheme: 'https', port: 443}, {body: 'body', 'status' => 200})`

By specifying host/scheme/port directly it avoids the URI parsing (which is where we ended up with the path issues before). This particular example say that anything matching that host/scheme/port should get that response (and since path is not specified, ANY path should get that response).

Conversely, you can use a regex here, but similarly I think you need to do it in the context of the parsed params (instead of the url). So in that case it might look like:

`Excon.stub({host: 'example.com', scheme: 'https', port: 443, path: /\d+/}, {body: 'body', 'status' => 200})`

That should match anything that matches the host/scheme/port and has a path containing one or more digits.

Does that answer your question?

Asdfif:
it works, thank you <3

[...]

PR #855, opened 2024-10-09
when creating mocks with empty path, default to slash to match request building

geemus:


geemus:
fixes #787

This closes the bug around trailing slashes from #787, though the docs could still use love.

PR #859, opened 2024-10-23
Fix changelog entry for v0.112

edmorley:
Follow-up to:
https://github.com/excon/excon/commit/0e06fb917992b127ad0eab029aa56d5874eab261


geemus (approved):
Oops, thanks for the fix!

geemus (approved):

PR #865, opened 2024-10-28
encoding aware query parsing

### LINKED ISSUE #863: Since 1.0.0, an escaped non-utf8 query parameter raises with webmock
Caused by https://github.com/excon/excon/pull/857

```rb
require "excon"
require "webmock"

include WebMock::API
WebMock.enable!

param = 'aäoöuü'.encode('iso-8859-1')
param = CGI.escape(param)
stub_request(:get, "https://example.com/?#{param}").to_return(body: "abc")

Excon.get("https://example.com/?#{param}")
```

This is only the case when using it with webmock, just with excon itself and no mocking it passes. The params are escaped, however I believe excon unescapes them here, https://github.com/excon/excon/blob/33a071d992f04350c12918a9b259154df35c5332/lib/excon/utils.rb#L77. This produces an UTF-8 string that is not valid in its encoding. This is then passed along to webmock, where certain operations (like regexp matching) fail.

geemus:
fixes #863

### DIFF
--- a/lib/excon/utils.rb
+++ b/lib/excon/utils.rb
@@ -74,13 +74,11 @@ module Excon
       params = {}
 
       string.split(/[&;] */n).each do |pair|
-        key, value = pair.split('=', 2).map { |x| CGI.unescape(x) }
+        key, value = pair.split('=', 2).map do |x|
+          URI.decode_www_form_component(x, string.encoding)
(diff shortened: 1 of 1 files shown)


geemus:
Trying a different tack elsewhere.

PR #867, opened 2024-10-29
Use casecmp? instead of casecmp() == 0

ekohl:
This has the same resulting value, but is easier to read. It's available in Ruby 2.7, which is currently the oldest required version.


adamruzicka (approved):

geemus (approved):
lgtm, thanks!

PR #868, opened 2024-10-29
Allow Excon::Connection to be instantiated without parameters

### LINKED ISSUE #866: Since 1.0.0, Excon::Connection cannot be instantiated without passing arguments to it
```
# ruby -e "require 'excon'; Excon::Connection.new"
/home/aruzicka/.asdf/installs/ruby/3.1.4/lib/ruby/gems/3.1.0/gems/excon-1.0.0/lib/excon/utils.rb:91:in `default_port?': undefined method `casecmp' for nil:NilClass (NoMethodError)

      (datum[:scheme].casecmp('http').zero? && datum[:port] == 80) ||
                     ^^^^^^^^
        from /home/aruzicka/.asdf/installs/ruby/3.1.4/lib/ruby/gems/3.1.0/gems/excon-1.0.0/lib/excon/utils.rb:65:in `port_string'
        from /home/aruzicka/.asdf/installs/ruby/3.1.4/lib/ruby/gems/3.1.0/gems/excon-1.0.0/lib/excon/connection.rb:103:in `initialize'
        from -e:1:in `new'
        from -e:1:in `<main>'
```

With excon-0.112.0 this passes.

adamruzicka:
Fixes #866

### DIFF
--- a/lib/excon/utils.rb
+++ b/lib/excon/utils.rb
@@ -87,8 +87,9 @@ module Excon
     end
 
     def default_port?(datum)
-      (datum[:scheme].casecmp?('http') && datum[:port] == 80) ||
-        (datum[:scheme].casecmp?('https') && datum[:port] == 443)
+      (!datum[:scheme]&.casecmp?('unix') && datum[:port].nil?) ||
+        (datum[:scheme]&.casecmp?('http') && datum[:port] == 80) ||
+        (datum[:scheme]&.casecmp?('https') && datum[:port] == 443)
     end
 
     def query_string(datum)
--- a/tests/connection_tests.rb
+++ b/tests/connection_tests.rb
@@ -91,5 +91,9 @@ Shindo.tests('Excon Connection') do
     end
   end
 
+  test('initialization') do
+    !Excon::Connection.new.nil? # This should not raise
+  end
+
   env_restore
 end


lib/excon/utils.rb
```diff
@@ -87,7 +87,8 @@ def parse_query_string(string)
     end
 
     def default_port?(datum)
-      (datum[:scheme].casecmp('http').zero? && datum[:port] == 80) ||
+      datum[:port].nil? ||
+        (datum[:scheme].casecmp('http').zero? && datum[:port] == 80) ||
```
ekohl, line 91: Don't you still want to make it safe if no scheme is known?
```suggestion
        (datum[:scheme]&.casecmp('http')&.zero? && datum[:port] == 80) ||
```

I'm also surprised that here the scheme is case insensitive but in other places the `HTTPS` constant is used.

https://github.com/excon/excon/blob/a8f625d1bd5eee8d99371b38751ac097562dcf23/lib/excon/connection.rb#L554

The constant is defined here:
https://github.com/excon/excon/blob/a8f625d1bd5eee8d99371b38751ac097562dcf23/lib/excon/constants.rb#L30

adamruzicka: > Don't you still want to make it safe if no scheme is known?

Done

> I'm also surprised that here the scheme is case insensitive but in other places the HTTPS constant is used.

It is odd, but it has been this way for over a decade, so I'd rather let it be

ekohl (approved):

geemus (approved):
lgtm, thanks!

lib/excon/utils.rb
geemus: I think it might be one of those "be generous with input, be strict with output" kind of things in terms of capitalization. But it's been so long I don't recall for sure. It might be worth thinking about more and cleaning up further, but this seems good-enough for now.

adamruzicka:
Rebased and resolved the conflict

tests/connection_tests.rb
```diff
@@ -91,5 +91,9 @@
     end
   end
 
+  test('initialization') do
+    Excon::Connection.new # This should not raise
```
ekohl, line 95: I think this is failing the tests because it's not `true`. How about:
```suggestion
    !Excon::Connection.new.nil? # This should not raise
```

ekohl (commented):
https://github.com/excon/excon/pull/867 was green, so we know CI is working.

tests/connection_tests.rb
adamruzicka: I can't get green tests locally even without my changes, so let's try that and we'll see

lib/excon/utils.rb
```diff
@@ -87,8 +87,9 @@ def parse_query_string(string)
     end
 
     def default_port?(datum)
-      (datum[:scheme].casecmp?('http') && datum[:port] == 80) ||
-        (datum[:scheme].casecmp?('https') && datum[:port] == 443)
+      datum[:port].nil? ||
```
geemus, line 90:
```suggestion
      (!datum[:scheme]&.casecmp?('unix') && datum[:port].nil?)
```

Besides the http/https cases, we also need to account for the `unix` scheme. This check should always return false for `unix`, because there isn't a default there. So to account for that, I think we need to add this extra guard that says if the schema is `unix` we want to return false, regardless of the port value. This appears to get my tests to green locally, fwiw. Make sense?

adamruzicka: Oh, I see. Changed

geemus (approved):

ekohl:
@geemus If this could be released it'd fix our CI, which I'd prefer over pinning excon.

geemus:
Released in v1.1.0

PR #869, opened 2024-10-29
change to using unencode based on uri addressable

### LINKED ISSUE #863: Since 1.0.0, an escaped non-utf8 query parameter raises with webmock

geemus:
fixes #863

### DIFF
--- a/lib/excon/utils.rb
+++ b/lib/excon/utils.rb
@@ -69,24 +69,15 @@ module Excon
       end
     end
 
-    # Used to normalize queries for stubs, based on Rack::Utils.parse_query
-    def parse_query_string(string)
-      params = {}
+    def split_query_string(string)
+      query = {}
 
       string.split(/[&;] */n).each do |pair|
-        key, value = pair.split('=', 2).map do |x|
-          # unescape, based on Addressable::URI unencode
-          x.gsub!(/%[0-9a-f]{2}/i) do |sequence|
-            c = sequence[1..3].to_i(16).chr
-            c.force_encoding(sequence.encoding)
-          end
-          x.force_encoding(Encoding::UTF_8)
-        end
-
-        params[key] = params[key] ? [params[key], value].flatten : value
+        key, value = pair.split('=', 2)
+        query[key] = query[key] ? [query[key], value].flatten : value
       end
 
-      params
+      query
     end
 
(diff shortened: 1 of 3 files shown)


lib/excon/utils.rb
```diff
@@ -74,13 +74,16 @@ def parse_query_string(string)
       params = {}
 
       string.split(/[&;] */n).each do |pair|
-        key, value = pair.split('=', 2).map { |x| CGI.unescape(x) }
+        key, value = pair.split('=', 2).map do |x|
+          # unescape, based on URI::Addressable unencode
```
sandstrom, line 78: Extremely minor thing, but just something I noticed.

```suggestion
          # unescape, based on Addressable::URI unencode
```

geemus: :facepalm: good catch. It's minor now, but I imagine it could save some major headaches in the future when we've all forgotten the original context/intent. Thanks!

geemus:
Seems like this is still failing to fix the test case in #863, need to think/dig a bit more.

geemus:
Also commented on #863, but will duplicate here to keep info together:
> Ok, I think I was overthinking/overdoing it. I've simplified https://github.com/excon/excon/pull/869 so that instead of trying to do escaping, it just splits based on &/; and :. That should get the original goal (normalizing so that if you use a hash to define the mock but a string to call it, or vice versa, it still matches). I think I was thinking about the real case (where escaping comes into play). Especially since this is all internal for the sake of matching stubs, the encoding wasn't really gaining us much even if it was working (and it wasn't working).

PR #871, opened 2024-10-31
use existing query_string to consistently cast query strings for mocks

geemus:


Earlopain:
Sorry for not responding on the other issues/PRs, I forgot.

I'm not sure if this PR or the one in webmock are the correct approach. There may be missing test coverage in webmock. How would webmock know if the hash was passed as is by the user or pushed into that shape by excon beforehand? When excon does the hash conversion, values are already escaped and webmock should not touch them but when it was passed as is by the user escaping must happen when reconstructing the query string.

All this to say, must the original `query` be overwritten or could you instead use something like `query_for_mock` to avoid overwriting what the user passed? If possible, I believe that would be the best apporach.

geemus:
@Earlopain good questions. It feels to me like the webmock implementation is making too many assumptions about what I would consider to be excon internals, unfortunately. Rather than setting a stub and then doing it's own handling, I think it might be better off having it's own middleware that was similar to the excon mock middleware, but had all it's own logic. That way we wouldn't step on each other with slightly different assumptions. Not sure about that. I do think this is a good improvement to things on this side either way though.

geemus:
Released in v1.2.0. Hope it will improve things, it shouldn't be any worse (though maybe comparably bad, unfortunately). We'll keep working on it.

PR #872, opened 2024-11-06
Include port for CONNECT when include_default_port is used

lcmen:
Thanks for this gem!

We're using http-based proxy (Fixie) to access secured URLs (https) and we stumbled upon an issue where version >= 1.0.0 is not able to read data - `EOFError` is raised.

I was able to pinpoint the problem to missing port in `CONNECT` directive. I've fixed the issue by respecting `include_default_port` when using `CONNECT` but I'm wondering if port should not be included there by default. WDYT?

I'd love to write a spec to cover this change but I'm not sure where to start. Would you be so kind to help me with that?

### DIFF
--- a/lib/excon/ssl_socket.rb
+++ b/lib/excon/ssl_socket.rb
@@ -108,7 +108,7 @@ module Excon
       end
 
       if @data[:proxy]
-        request = "CONNECT #{@data[:host]}#{Excon::HTTP_1_1}" \
+        request = "CONNECT #{@data[:host]}#{port_string(@data)}#{Excon::HTTP_1_1}" \
                   "Host: #{@data[:host]}#{port_string(@data)}#{Excon::CR_NL}"
 
         if @data[:proxy].has_key?(:user) || @data[:proxy].has_key?(:password)


geemus:
Oops, seems like a pretty obvious oversight now that you point it out. I don't use proxies much in my day to day, so mostly relied on the test suite, which obviously has a gap as you suggest.

In general I'd say our proxy tests could use some love. In no small part because it's not something I make much use of, I haven't had the highest confidence. That being said, I think part of the sticking point had been how we would run a proxy for the tests in particular. I think at the time at least, I didn't find great options. Looking again now, Webrick has an http proxy that might be promising. ie maybe we could run commands against it which modified the request/response with a handler. We could then test that the request/responses worked (and were modified and expected). I think that at least would be much more thorough than what we have now, though I don't know for sure if it would have caught this particular bug.

Here is the proxy I found, fwiw: https://github.com/ruby/webrick/blob/cb93ef8144b10832b591a7e8c472cdd5d024a750/lib/webrick/httpproxy.rb#L69. I think we already use webrick in the test suite, so it would be available if I recall properly.

Would you be interested in collaborating on trying to get that working? We could probably also include your change and follow up with the tests, it's certainly no-worse testing wise than where we were and fixes a bug. What do you think?

lcmen:
Thank you @geemus for quick response and all the suggestions for implementing proxy. I will need some time to investigate how I could include Webrick's proxy (or any other solution) for testing.

Would you be open to merge this PR and add tests as a follow-up? At the moment, we had to rollback to 0.112 on our `production`.

One more question about the error itself. We were getting `EOFError` and when I looked more carefuly it seemd like `@socket` contained an empty response so excon could not read from it. I'm wondering if we could somehow add a guard or use different error class it so it's easier to understand the reason for failure. WDYT?

geemus:
@lcmen Yeah, that works for me. Any help on the tests around proxies would be greatly appreciated when you get a chance (and it doesn't need to be webrick, that's just one option I found, fwiw). I'll bring this in and see about a release to fix the immediate issue.

geemus:
Released in v1.2.0

lcmen:
Thank you @geemus for merging and releasing new version so quickly.

I will look into a way to bring Webrick proxy (I think it should work) into specs. First, I need to write my head around the spec setup as I see the migration from `tests` -> `specs` is still in progress. I also see that the mocking code for servers (e.g. for spinning up the rack processes is there). Should I follow that convention for proxy as well?

lcmen:
Thanks @geemus ! I will see what I can do.

Before that one extra question related to improving the proxy setup: should we always include port for `CONNECT no matter what `include_default_port` option is?

The more I think about it, the more I wonder about the purpose of having `include_default_port` in the first place. What's wrong with always including the port, even for default ones (like it was with version < 1.0.0)?

geemus:
@lcmen I'm not sure if we should disregard that option or not. If I recall correctly, the RFCs suggest that the port should be optional (so things should probably work either way). It should match between the headers and the initial part of the request (so the fix here was important), but beyond that it should be fine. In practice, however, we saw lots of weird issues where in the wild this just wasn't true. So providing this option was a release valve for cases where peoples setup/configuration don't match with the spec. I'm not sure if the CONNECT case is special/different in this regard, but I'd lean toward keeping the release valve around at least for a bit longer. I'll also note that prior to 1.0, default port was always included (unless you opted out) and we reversed that behavior in 1.0. We haven't yet seen how that will shake out, though I suspect it will be less likely to cause issues. Anyway, my inclination remains one of "wait and see". Removing it would be great if we find it isn't needed and can simplify, but I'm not quite read yet.

PR #874, opened 2024-11-14
Handle read error from socket after "Broken pipe"

### LINKED ISSUE #873: There is not always data to read from socket when `Errno::EPIPE` is raised (on macOS?)
In #761, Excon started to read the server response from the socket. However, there are cases where a `Errno::EPIPE` is raised and the socket is not readable. I have only been able to reproduce it under macOS (currently 14.7.1) and apologies in advance for the clunky repro:

```
# on Ruby 3.3.5 and Excon 1.2.0
$ ruby -rexcon -e 'p RUBY_VERSION;p Excon::VERSION'
"3.3.5"
"1.2.0"

# start a webserver
$ ruby -run -e httpd . -p 8000

# make a persistent connection and sleep for a while before making a new request, during the sleep restart the webserver in the step above
$ ruby -rexcon -e 'conn = Excon.new("http://localhost:8000", persistent: true); conn.get; sleep 70; conn.get(idempotent: true)'
<internal:io>:63:in `read_nonblock': Connection reset by peer (Errno::ECONNRESET)
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/socket.rb:249:in `read_nonblock'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/socket.rb:67:in `read'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:197:in `rescue in request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:115:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/mock.rb:57:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/instrumentor.rb:34:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/idempotent.rb:19:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/base.rb:22:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/decompress.rb:14:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/base.rb:22:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:293:in `request'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:379:in `get'
  from -e:1:in `<main>'
<internal:io>:121:in `write_nonblock': Broken pipe (Errno::EPIPE)
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/socket.rb:321:in `block in write_nonblock'
  from <internal:kernel>:187:in `loop'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/socket.rb:316:in `write_nonblock'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/socket.rb:106:in `write'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:167:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/mock.rb:57:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/instrumentor.rb:34:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/idempotent.rb:19:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/base.rb:22:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/decompress.rb:14:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/middlewares/base.rb:22:in `request_call'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:293:in `request'
  from <...>/3.3.5/lib/ruby/gems/3.3.0/gems/excon-1.2.0/lib/excon/connection.rb:379:in `get'
  from -e:1:in `<main>'

# paths edited for brevity
```

Following the stack and one can see that it tries to read from the socket at https://github.com/excon/excon/blob/v1.2.0/lib/excon/connection.rb#L197, causing the `Errno::ECONNRESET` to be raised, but not wrapped as a `Excon::Error::Socket` as intended. Ultimately leading to the "idempotent retry" to never happen.

I have found that the needed sleep is at least 60 seconds in order for this to happen. I first thought it was related to the timeout settings, I tried to decrease them but it does not seem to change anything.

Interestingly enough, under Linux I get a `EOFError` rather than `Errno::EPIPE` as the first raised error in the same scenario, and then retries works just fine. Moreover we have only observed this in local development (under macOS), production runs on Linux where we have never seen this happen, further making me believe this is isolated to macOS.

I can also add that if using `https`, the final raised error will be `OpenSSL::SSL::SSLError` (rather than `Errno::ECONNRESET`). I would assume this is because it uses `Excon::SSLSocket` instead of `Excon::Socket`.

I am not familiar with Excon to see what the proper solution would be, a `rescue` for the raise at https://github.com/excon/excon/blob/v1.2.0/lib/excon/connection.rb#L197 would be simple, but perhaps this should be caught at a deeper level.

Let me know if you need more input and thanks for the gem!

walro:
Under certain circumstances it's not possible to read from the socket, this ensures we still propagate a `Excon::Error` even if `Excon::Socket#read` raises.

As the reproduction is a bit tricky (see #873) I am not sure how to design the test. First we need a `Errno::EPIPE` and then we also need it to error out on the socket read.

In general I try to avoid one-line rescues as they catch `StandardError` (very broad). In this case I think it's fine though, I have seen `Errno::ECONNRESET` and `OpenSSL::SSL::SSLError` being raised here and there are probably more things that could happen. Moreover, we are already quite nested and adding yet another begin/end block with explicit rescues may not be too aesthetically pleasing :). I also contemplated adding a method, something like `try_read_from_socket(socket)` but didn't feel it was worth the indirection.

Close #873.

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -194,7 +194,7 @@ module Excon
           raise(error)
         when Errno::EPIPE
           # Read whatever remains in the pipe to aid in debugging
-          response = socket.read
+          response = socket.read rescue ""
           error = Excon::Error.new(response + error.message)
           raise_socket_error(error)
         else


geemus (approved):

geemus:
LGTM, thanks!

Agreed that this is not like what I would probably do in general/elsewhere, but that it seems fine for this narrow context. Especially since it basically immediately re-raises and would only alter the contents of the error message that ultimately results.

Similarly, I'm not sure how best to test this (and even if we figured it out it seems like it would potentially be overly specific and/or brittle).

All that said, it also more or less allows for the old behavior of this code (prior to when it tried to read from the socket to aide debugging), which makes it feel probably reasonably safe.

geemus:
Released in v1.2.1.

PR #875, opened 2024-11-21
Add funding_uri to gemspec

andrew:
I noticed that the project already has a FUNDING.yml, this PR adds `funding_uri` to the gemspec to help increase visibility using the `bundle fund` command in Bundler.


geemus (approved):
Thanks, I had noticed this was a thing, but wasn't sure how annoying it would be and hadn't gotten around to digging deeper. Seems fine to try out and easy enough to back out if we change our minds.

andrew:
I think by default it just shows a count of the number of gems that have a funding link when you run bundle install, rather than just blurting out a massive list to everyone, which is nice.

PR #877, opened 2025-01-18
Don't use rescue in modifier form with condition

Earlopain:
This is actually parsed as `socket.close rescue (nil if socket)`. In practise this doesn't matter since it already swallows all errors, including the `NoMethodError` when the socket is actually nil.


geemus (approved):
LGTM, thanks!

geemus:
Released in v1.2.4

PR #878, opened 2025-02-21
Add ignore EOF constant to ignore OpenSSL EOF error

Manisha15:
We have this issue in Proxmox https://github.com/theforeman/foreman_fog_proxmox/issues/325, where Proxmox has open SSL connection which is not closed and results in  SSL routines::unexpected eof. This PR adds new constant which when set true ignores unexpected EOF in OpenSSL connection.

### DIFF
--- a/lib/excon/constants.rb
+++ b/lib/excon/constants.rb
@@ -86,6 +86,7 @@ module Excon
     keepalive
     host
     hostname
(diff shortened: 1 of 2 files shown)


geemus (approved):
lgtm, thanks!

geemus:
Released in v1.2.4

PR #879, opened 2025-02-23
Basic request works inside a Ractor

betelgeuse:
Do a lot of constant freezing so the library can make a simple request inside a ractor.

Resolv::DefaultResolver is defined by just calling new. The overhead should be acceptable.


betelgeuse:
> Hey, thanks for taking a pass at this. Looks like the parallel test/spec stuff bites us again, in that we are getting failures from test_helper (even though you updated the spec_helper). Could you look at those and see about making similar changes there as well?

Added. The non rspec tests didn't pass for me locally with the changes but it's unclear whether it's related to the changes. Need to test with vanilla master to see if these are the reason for the breakage.

geemus:
@betelgeuse sounds good, let me know if you have questions or need help. The doubled test suite is a weird historical artifact in a lot of ways, but also would be a lot of work to "fix", so it's been kind of stuck that way for a while.

betelgeuse:
> @betelgeuse sounds good, let me know if you have questions or need help. The doubled test suite is a weird historical artifact in a lot of ways, but also would be a lot of work to "fix", so it's been kind of stuck that way for a while.

CI seems to have passed except on jruby so MRI is good. Is this jruby failure new?

geemus:
@betelgeuse Unfortunately not, jruby has been broken here for some time (it's why it's allowed to fail in the config). The key thing we are looking for is passing MRI.

geemus:
Were you able to get things passing locally?

betelgeuse:
> Were you able to get things passing locally?

No. But this also fails locally with master:

```
  Request Tests
    persistent connections
      with default :persistent => true, 127.0.0.1:9292
        + uses a persistent connection
        + :persistent => false resets connection
      with default :persistent => false, 127.0.0.1:9292
        + does not use a persistent connection
        + :persistent => true enables persistence
      undefined method `include?' for nil:NilClass (NoMethodError)
        tests/test_helper.rb:392:in `with_server'
        tests/request_tests.rb:6:in `block (3 levels) in <top (required)>'
        tests/request_tests.rb:5:in `each'
        tests/request_tests.rb:5:in `block (2 levels) in <top (required)>'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:79:in `instance_eval'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:79:in `tests'
        tests/request_tests.rb:2:in `block in <top (required)>'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:79:in `instance_eval'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:79:in `tests'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:38:in `initialize'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:13:in `new'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo.rb:13:in `tests'
        tests/request_tests.rb:1:in `<top (required)>'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo/bin.rb:61:in `load'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo/bin.rb:61:in `block (2 levels) in run_in_thread'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo/bin.rb:58:in `each'
        /var/lib/gems/3.2.0/gems/shindo-0.3.11/lib/shindo/bin.rb:58:in `block in run_in_thread'
```

geemus:
@betelgeuse thanks for the context. Probably good to fix (even though it doesn't block this PR).

Looks like the line in question is: `until (line = e.gets).include?('ready')`. I thought e.gets would block in that context until it actually had something (so I don't know how it's getting a nil).

That said, maybe we should add a safe accessor. So something like: `until (line = e.gets)&.include?('ready')`.

Could you try that and see if you get any further?

geemus:
@betelgeuse Do you know if newer ruby DNS is ractor compatible? Also, do you need a release with these changes?

betelgeuse:
> @betelgeuse thanks for the context. Probably good to fix (even though it doesn't block this PR).
> 
> Looks like the line in question is: `until (line = e.gets).include?('ready')`. I thought e.gets would block in that context until it actually had something (so I don't know how it's getting a nil).
> 

My working theory is that it's returning nil because nothing is ever written to the stream and it's closed.

```Returns nil if called at end of file.```

https://www.apidock.com/ruby/IO/gets

geemus:
@betelgeuse ah, good call, thanks for pointing that out. Happy to work on debugging that further if you want, but I totally understand if you are satisfied for now with this merged. Master seems fine for me (ruby 3.2.2 it looks like) locally, and I hadn't heard of that particular issue before. So I'm not exactly sure why you would end up with that never writing and ultimately closing (or whatever it's doing, that seems like a good theory though).

PR #883, opened 2025-03-01
Replace magic number with signal name

betelgeuse:
It's easier to read code without magic numbers.

### DIFF
--- a/benchmarks/excon.rb
+++ b/benchmarks/excon.rb
@@ -40,7 +40,7 @@ def with_server(&block)
   end
   yield
 ensure
-  Process.kill(9, pid)
+  Process.kill('KILL', pid)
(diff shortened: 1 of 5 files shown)


betelgeuse:
@geemus there seems to be a lot of duplication around process handling. Is it something that should be cleaned? Not promising I will do the refactoring but in case I do asking before spending time on it.

geemus (approved):
lgtm, thanks!

geemus:
@betelgeuse - thanks for the help on things. I'm definitely open to help cleaning things up. I think the process stuff has often been something that was worked on just enough to solve whatever the immediate concern was before moving on to other things, so I'm not surprised that it could use some work.

PR #900, opened 2026-05-14
Raise Excon::Error::InvalidParameter when host is nil

### LINKED ISSUE #899: NoMethodError: undefined method '+' for nil when host is nil
### Summary

When `datum[:host]` is nil at request time, Excon crashes with a cryptic `NoMethodError` instead of raising a descriptive `Excon::Error`. The crash site is `connection.rb:257`:

```ruby
datum[:headers][host_key] ||= datum[:host] + port_string(datum)
#                               ^^^^^^^^^^^^ nil → NoMethodError: undefined method '+' for nil
```

### Ways to trigger a nil host

**1. URL with no authority component**

```ruby
Excon.get("https:/registry.example.com/package")  # single slash — no "//"
Excon.get("https:package")                         # no slashes at all
```

`URI.parse` returns `nil` for `.host` on both of these.

**2. Explicit `host: nil` option overriding the parsed host**

```ruby
Excon.get("https://registry.example.com/package", host: nil)
# {host: "registry.example.com"}.merge({host: nil}) → {host: nil}
```

**3. `Excon::Connection.new` built directly without a host**

```ruby
Excon::Connection.new(scheme: "https", path: "/package", port: 443).get
```

### Reproducer

Full reproducer with CI covering all three cases: https://github.com/yeikel/reproducer-excon-issue-899

### Expected behaviour

A descriptive `Excon::Error` should be raised when the host is nil, rather than letting Ruby crash with a `NoMethodError` that gives no indication of what went wrong.

### Suggested fix

Raise a descriptive error before the concatenation:

```ruby
raise Excon::Error, "host is nil — the URL may be malformed" if datum[:host].nil?
datum[:headers][host_key] ||= datum[:host] + port_string(datum)
```

### Environment

- Excon: all versions tested (1.3.2, 1.4.2, latest)
- Ruby: 3.1, 3.2.11, 3.3, 3.4, 4.0.4

yeikel:
## Summary

- Adds `Excon::Error::InvalidParameter` to the error hierarchy (alongside `InvalidHeaderKey`/`InvalidHeaderValue`)
- Raises it early in `Connection#initialize` when `:host` is explicitly set to `nil` (malformed URL, explicit `host: nil` override)
- Raises it in `Connection#request` when `datum[:host]` is nil at request time (connection built without a host key)
- Adds 3 tests covering all trigger paths from issue #899

## Why `Excon::Error::InvalidParameter` and not `ArgumentError`

Users rescue `Excon::Error` to catch library failures. An `ArgumentError` would silently escape that rescue block, making the nil-host case impossible to handle consistently alongside network errors.

## Test plan

- [x] `Excon.new('https:/registry.example.com/package')` → raises `Excon::Error::InvalidParameter`
- [x] `Excon.new('https://example.com', host: nil)` → raises `Excon::Error::InvalidParameter`
- [x] `Excon::Connection.new(scheme: 'https', path: '/package', port: 443).get` → raises `Excon::Error::InvalidParameter`
- [x] `Excon::Connection.new` (no args) → still does not raise (existing `initialization` test preserved)
- [x] Full test suite: `12 pending and 1471 succeeded`, no failures

Closes #899

### DIFF
--- a/lib/excon/connection.rb
+++ b/lib/excon/connection.rb
@@ -102,6 +102,10 @@ module Excon
           @socket_key = "#{@data[:scheme]}://#{@data[:socket]}"
         end
       else
+        if @data.key?(:host) && @data[:host].nil?
+          raise Excon::Error::InvalidParameter, "host is required for non-Unix connections; " \
+                                                "the URL may be malformed (e.g. missing \"//\" after the scheme)"
+        end
         @socket_key = "#{@data[:scheme]}://#{@data[:host]}#{port_string(@data)}"
       end
       reset
@@ -256,6 +260,10 @@ module Excon
       if datum[:scheme] == UNIX
         datum[:headers][host_key] ||= ''
       else
+        if datum[:host].nil?
+          raise Excon::Error::InvalidParameter, "host is required for non-Unix connections; " \
+                                                "the URL may be malformed (e.g. missing \"//\" after the scheme)"
+        end
         datum[:headers][host_key] ||= datum[:host] + port_string(datum)
       end
 
--- a/lib/excon/error.rb
+++ b/lib/excon/error.rb
@@ -48,6 +48,7 @@ or:
 
     class InvalidHeaderKey < Error; end
     class InvalidHeaderValue < Error; end
+    class InvalidParameter < Error; end
     class Timeout < Error; end
     class ResponseParse < Error; end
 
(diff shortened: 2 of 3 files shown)


geemus (approved):
lgtm, thanks for the detailed report and the fix. I hope to cut a release including this and some other fixes a bit later today.

PR #903, opened 2026-07-13
Introduce read_buffer_scratch

kukicola:
Hello
Recently, I upgraded one of the apps to Ruby 4 and instantly noticed a huge memory increase. After digging, I found out that for some reason, strings created by reading data from sockets without a buffer are not being cleared by the GC (no idea what changed in Ruby 3->4). I fixed the issue by adding an additional buffer that is used when the main one is not empty, instead of creating new objects on every read. In my case, it reduced memory usage by 30% on instances that use excon to ingest data from APIs.

Env: Ruby 4.0.5, Alpine 3.24.1, jemalloc 5.3.0

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -52,7 +52,8 @@ module Excon
       @data = data
       @nonblock = data[:nonblock]
       @port ||= @data[:port] || 80
-      @read_buffer = String.new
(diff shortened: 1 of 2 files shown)


geemus (approved):
Thanks, lgtm.

geemus:
released in v1.6.0

PR #904, opened 2026-07-15
Fix incorrect variable name

cedricpim:
Avoid a NameError but properly passing variable "e" instead of non-existing "error".

### DIFF
--- a/lib/excon/socket.rb
+++ b/lib/excon/socket.rb
@@ -298,7 +298,7 @@ module Excon
     rescue OpenSSL::SSL::SSLError => e
       select_with_timeout(@socket, :read) && retry if e.message == 'read would block'
 
-      raise(error)
+      raise(e)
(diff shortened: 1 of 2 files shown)


azamb:
@cedricpim ty for putting this PR together. I think a similar issue exists in [`write_nonblock`](https://github.com/excon/excon/blob/30167a41a8ac8baf1b00c9f6e8f8e168da0b9984/lib/excon/socket.rb#L308-L328)


the rescue block starting on line 324 should reference `e` instead of `error`. Do you wanna include that change as part of the scope of this PR? Otherwise I can submit a separate one.

geemus (approved):
lgtm, thanks!

geemus:
@azamb quick fix for the one you spotted in #905

azamb:
@geemus thank you for the quick turn around! Any chance of a new minor release being published with these fixes?

geemus:
Both released in v1.6.0

Issue #651, opened 2017-10-27
Memory leak when using short lived threads


alan:
In the last few days I've been chasing a memory leak that was introduced as part of the fix of #640. The memory leak of the fix is actually discussed in the conversation about the fix and we've run into it.

We use Excon to keep a persistent connection to an API in a web application that runs with Puma as the server. The Puma configuration allowed for a variable number of threads between a minimum and maximum (this is how Puma runs by default). With the new change the app uses all of the memory on the server and had to be killed.

Before the change the sockets for each connection were cached on the thread variables, so when a Puma thread got destroyed it allowed the Excon sockets to be garbage collected. With the new version the cache is stored on the connection itself. When a thread is destroyed the cache does not clear the sockets that it holds for that thread which causes a leak of Excon sockets.

I'm not sure what is the preferred fix, there are a couple of solutions that I can think of:

- The first one would be removing the cache. I found it surprising and unexpected that sockets are shared between connections, something also discussed in #640. (In our case to ensure thread safety we use a connection pool of Excon connections that the Puma threads can access)

- Use [ObjectSpace#define_finalizer](http://ruby-doc.org/core-2.4.2/ObjectSpace.html#method-c-define_finalizer) to cleanup the socket caches when a thread is destroyed.

I prefer the first option because is simpler and leads to less confusion about how an Excon connection really works. The second solution seems a bit hackish to me.

However I may be missing a use case where the cache is necessary or useful?

geemus:
Thanks for the detailed report. Maybe it is in fact time to pull the plug on the "automagic" caching across connections and limit it to just caching within the context of a particular connection. I believe that is what 1 means, correct?

The idea with the cache was that it would give you a better default behavior even if you were lazy about reusing connections (because initially at least, people were not managing connection pools/etc). That is much more a nice-to-have than a requirement though, especially now that we have seen clear downsides for advanced users that seem to greatly outweigh the potential nominal benefit for beginners. Less advanced usage should still work in this case, though it might be slightly less performant. And there is a clear path to improved performance as well. As such, dropping it seems safe/reasonable enough to me.

As a bit more of an aside, to take this to the next level, I do sometimes wonder if we shouldn't consider providing some connection pooling functionality as part of the gem in order to make it easier for advanced users to do advanced things. It may be something where peoples needs differ enough to make it hard for a more general case solution, but I would be interested to hear more about how you are managing your pool and how that's working for you (this issue aside).

[...]

PR #906, opened 2026-07-18
Use weak keys for thread-safe socket caches

kukicola:
Hello again,
I have another memory optimization change that would solve https://github.com/excon/excon/issues/651. Using `WeakKeyMap` would allow the GC to clear sockets for threads that no longer exist. Unfortunately, it was introduced in Ruby 3.3. I'm not sure if it's ok to bump the required version, although 3.2 already reached EOL


geemus:
Released in v1.7.0.

